PR #27, opened 2011-03-17
Ability to change PEP8 flags on the fly in source file via inline comments

smira:
Sometimes we need to violate PEP8 rules to improve readability of source file, at the same time we would like file to be still checked by PEP8.

I've added simple change that allows to influence pep8 'expected' codes:
# :PEP8 -E221 -W603

a    = 3
long = 4
# :PEP8 +E221

Prepending minus sign disables the code, while prepending plus enables it back.

Please let me know if that makes sense for inclusion, I may need to add tests...

### DIFF
--- a/pep8.py
+++ b/pep8.py
@@ -1215,11 +1215,13 @@ class Checker(object):
             #:PEP8 +E101 -W603
             """
             return set(map(lambda s: s[1:].strip(),
-                filter(lambda s: s.startswith(sign), line.split(" ")[1:])))
+                           filter(lambda s: s.startswith(sign),
+                                  line.split(" ")[1:])))
 
         if line.lstrip().startswith('#:PEP8'):
-            self.expected = (self.expected | parse_inline_flags('-')) \
-                    - parse_inline_flags('+')
+            self.report.expected = \
+                (self.report.expected | parse_inline_flags('-')) \
+                - parse_inline_flags('+')
         if line:
             self.check_physical(line)
         return line
@@ -1328,7 +1330,6 @@ class Checker(object):
         """
         Run all checks on the input file.
         """
-        self.expected = set(expected or ())
         self.line_offset = line_offset
         self.report.init_file(self.filename, self.lines, expected, line_offset)
         self.line_number = 0
@@ -1400,7 +1401,7 @@ class BaseReport(object):
         """Signal a new file."""
         self.filename = filename
         self.lines = lines
-        self.expected = expected or ()
+        self.expected = set(expected or ())
         self.line_offset = line_offset
         self.file_errors = 0
         self.counters['files'] += 1


apeyser:
Wouldn't a better solution be to make pep8 configurable to respect file level suppressions? Then everyone gets what they want. The suggestion to fork the project seems a bit extreme, when you can easily set a local policy to either use or not use, since the use of pep8 itself is a policy and not a technical decision. The complexity is already there, insofar as it's possible to create pep8 plugins, it's just a question of where that complexity is exposed.

Otherwise, using automated pep8 compliance becomes a problem --- you have to hack into your infrastructure exceptions here and exceptions there. The current flexibility is sufficient for most projects which use pep8 in the manner that they use them currently --- by definition. A proliferation of non-standard hacks distributed in different projects would seem suboptimal (see the comment above).

Maybe the solution is a project for a plugin that then goes into pip, and have in pep8 documentation a link to that project so that people don't come back adding issues and necrobumps (as I am)?

sigmavirus24:
> Wouldn't a better solution be to make pep8 configurable to respect file level suppressions?

As in something that happens once in a file and tells flake8 what to ignore in that file? No. We've furiously avoided that as best as we can. That introduces a level of complexity that no one really wants to maintain.

> it's possible to create pep8 plugins

Actually only Flake8 allows you to register plugins. Flake8 does quite a bit of work with pep8 to get it to accept our plugins.

> Maybe the solution is a project for a plugin that then goes into pip, and have in pep8 documentation a link to that project so that people don't come back adding issues and necrobumps (as I am)?

I have no clue what you're asking about a project going into pip. Regardless, to implement per-file supressions would require per-file StyleGuides which would then flip the way pep8 works on its head. Hence the need for a fork. I'm sorry, but the way pep8 has been architected does not allow for this since it applies a style guide to the entirety of a run rather than a style guide per file. (Typically projects have a single style guide, not N where N is the number of files in a project.)

apeyser:
After looking over the code, a new project would probably be best for this for a variety of reasons. In fact, it would probably also be much better for flake8 as well to use a tool with a different interface (thus problems with needing to pin versions it seems from the flake8 docs). Like everyone else, if I find the time to write such a module from scratch...

Typically a project has one style guide for most code, but there's always a need in a large enough project for sets with specific didactic needs and thus an avoidance of the standard guide, technical clarity reasons for specific modules to not follow the global style guide, and other good technical or social reasons for variance. Consistency, hobgoblins and such.

emreay-:
@florentx 
> Personally, I don't feel comfortable with this noise in source file, which has no added value for the reader.
I prefer to have some false positive in the PEP8 report in such case, than to add this kind of comments.

One's preferences or whatever they feel comfortable should be irrelevant in any software tool (library/framework/linter and so on). You cannot easily make assumptions on your client's use cases and imposing personal preferences as opposed to providing flexibility is certainly an anti-pattern.

Plus the argument of "noise in source file" contradicts with the decision being made. There can be very justifiable cases (as in the first message) where the amount of noise could be a single line at the top. But now there is no way of doing that and each line that needs ignoring will have noise. Which is obviously more noise at the end.

Disappointing decisions.

sigmavirus24:
@emreay- trying to revive a conversation last commented on in 2016 is not productive nor is trying to shame a developer who is no longer part of the project.

Further behaviour like this will not be tolerated

PR #272, opened 2014-04-09
Fix the bug of diff mode: Don't check the context

zhiming-aa:
The diff mode should not include the context.


IanLee1521:
This has become very out of sync. If this is still an issue, please squash rebase it and open a new issue.

PR #287, opened 2014-05-09
Fix 'variable is referenced before assignment' error

dreadatour:
I think it is just mistype. It's works well without this small fix, but looks a little bit ugly.


IanLee1521:
Thanks for the Pull Request! I modified it slightly, and wasn't able to use your commits directly (repo was deleted maybe?), but it's made it in there!

dreadatour:
Oh, sorry for that. Thank you!

PR #294, opened 2014-05-28
Adds --valid-hangs to choose valid hanging indent

remram44:
This option allows you to set specific indentation length(s) that override the default `[4]` or `[4, 8]`, for example `--valid-hangs 4` or `--valid-hangs 3,17` (if you're insane).

PEP8 deprecates the use of tabs, but states that "hanging indents *may* be indented to other than 4 spaces".

This might help with some tab use-cases, and also with the very common use of 8 spaces for hanging indent in standard statements (since this is recommended for `if: with: def:` blocks).

Means to replace #291 (which only added `--double-indent`, equivalent to `--valid-hangs 8`).

### DIFF
--- a/pep8.py
+++ b/pep8.py
@@ -370,7 +370,7 @@ def indentation(logical_line, previous_logical, indent_char,
 
 
 def continued_indentation(logical_line, tokens, indent_level, hang_closing,
-                          indent_char, noqa, verbose):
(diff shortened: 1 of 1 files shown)


IanLee1521:
Since there hasn't been any other interest in this being a native part of pycodestyle, I'm going to go ahead and close this. Also, flake8 is probably a better place to add extensions like you're describing.

PR #304, opened 2014-07-24
Report E402 for imports not on top of file; issue #264. Also add ability...

yole:
... to keep local state for each checker plugin. #ep14boat


renstrom:
Won't this report false positives for projects using monkey patching in Gevent?

```python
from gevent.monkey import patch_all
patch_all()

import other  #  E402 violation
```

yole:
Looks like gevent monkey patching pretty much forces you to violate PEP 8. The diagnostics in this case is not a false positive; it's working as intended.

Issue #197, opened 2013-05-14
 The preferred place to break around a binary operator is after the operator, not before it


elopio:
Pep8 recommends to break a long line after a binary operator.
This library doesn't catch when we break before the operator.

IanLee1521:
Closed with Pull #305 . Thanks for the report!

larsyencken:
I think this is a misreading of PEP8. The PEP allows both, but weakly recommends the opposite behaviour, suggesting Knuth's style of breaking _before the operator_ for new code.

From the PEP:

```python
# No: operators sit far away from their operands
income = (gross_wages +
          taxable_interest +
          (dividends - qualified_dividends) -
          ira_deduction -
          student_loan_interest)

# Yes: easy to match operators with operands
income = (gross_wages
          + taxable_interest
          + (dividends - qualified_dividends)
          - ira_deduction
          - student_loan_interest)
```

https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator

I can disable this locally for myself, but I don't think this one should be turned on by default.

larsyencken:
Sorry for commenting on a very old issue. Looks like PEP8 has have been updated, there was more discussion about this in #498.

IanLee1521:
@larsyencken -- No worries! That's one of the difficulties of this tool, it's trying to hit a moving target, as the actual PEP is updated from time to time.

robvdl:
I can't get W504 passing after this fix and recent pycodestyle release.

This isn't passing:

        expected_url = (CONFIG['client']['url_prefix']
                                 + CONFIG['client']['url_path']
                                 + CONFIG['client']['create_parcel'].format(service_code='runjump'))

Neither is this:

        expected_url = (CONFIG['client']['url_prefix'] +
                                 CONFIG['client']['url_path'] +
                                 CONFIG['client']['create_parcel'].format(service_code='runjump'))

Is there possibly a bug in the latest pycodestyle release, or can someone tell me what I am doing wrong, in the mean time I'm having to disable W504 as this is driving me pretty crazy :)

sigmavirus24:
@robvdl "recent pycodestyle releases" is unfortunately vague. Can you please list which versions you've tried and your configuration?

sigmavirus24:
[...]

PR #305, opened 2014-07-24
Report W503 for line breaks before binary operators; issue #197. #ep14boat

yole:


IanLee1521:
Thanks for the patch!

schmiddy:
Hi, it seems that PEP8 has [recently been updated](http://www.mail-archive.com/python-dev@python.org/msg91905.html) to endorse breaking lines **before** a binary operator, and to suggest that this way be preferred for new code.

I propose that either:
 * W503 go away entirely, or
 * warning should only be issued when a module uses conflicting styles (i.e. breaking lines both before and after binary operators)

The-Compiler:
@schmiddy see #498

PR #313, opened 2014-08-08
allow use # noqa to disable E721 warning

hongqn:
Sometimes `isinstance()` is not suitable for comparing types.  For
example, someone may want to make sure an object is exact instance of a
type, not instance of its subclasses.

This patch allows to use `# noqa` comment to disable warning in this case.

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -323,7 +323,7 @@ This is the current list of error and warning codes:
 +----------+----------------------------------------------------------------------+
 | E714     | test for object identity should be 'is not'                          |
 +----------+----------------------------------------------------------------------+
-| E721     | do not compare types, use 'isinstance()'                             |
+| E721 (^) | do not compare types, use 'isinstance()'                             |
 +----------+----------------------------------------------------------------------+
 | E731     | do not assign a lambda expression, use a def                         |
 +----------+----------------------------------------------------------------------+
--- a/pep8.py
+++ b/pep8.py
@@ -974,7 +974,7 @@ def comparison_negative(logical_line):
             yield pos, "E714 test for object identity should be 'is not'"
 
 
-def comparison_type(logical_line):
+def comparison_type(logical_line, noqa):
     r"""Object type comparisons should always use isinstance().
 
     Do not compare types directly.
@@ -990,7 +990,7 @@ def comparison_type(logical_line):
     Okay: if type(a1) is type(b1):
     """
     match = COMPARE_TYPE_REGEX.search(logical_line)
-    if match:
+    if match and not noqa:
         inst = match.group(1)
         if inst and isidentifier(inst) and inst not in SINGLETONS:
             return  # Allow comparison for types which are not obvious


IanLee1521:
@hongqn -- 

I'm going to go ahead and close this for now. If you think this is a true issue, I might suggest that you look into changing the way that the PEP8 document specifies this. But I agree with @sigmavirus24 that is strongly advised against in the document.

hongqn:
Sorry for the late response.

The situation I need for `# noqa` is that I was implementing a serialization function for my project.  For some reason, it only accept predefined types.  Instances of subclass of these types are not acceptable.  So I must check if the type of given object is exactly one of the known types.

@IanLee1521 's comment said `type(a) is int` is PEP8 clean, this is odd.  I made some tests as the following:

```python
import types
from types import IntType
a = 1
type(a) is int  # no warning
type(a) is types.IntType  # E721
type(a) is IntType  # no warning
```

The last two lines should be identical but it seems pep8 tool treats them differently.  I'd like to consider it a bug to be inconsistent.

I checked the source code of pickle.py of Python 2.7.9 and found there are type comparation, too.  Like these:

```
% grep -n 'type(.* is'  pickle.py
316:        if type(rv) is StringType:
321:        if type(rv) is not TupleType:
1048:                type(klass) is ClassType and
```

It passed pep8 check only because it uses `from types import *` instead of `import types`.

So, what do you think which is preferred for a serializer developer, to use a `# noqa` to tell pep8 checker to ignore that line, or to use `from types import xxx` as a workaround?  I think the latter more like a "depend on implementation bug, not feature".

IanLee1521:
Based on some emails with @ncoghlan I'm going to go ahead and merge this PR. That said you should still probably try to avoid using the ``# noqa`` option when possible.

IanLee1521:
For the record, based on the emails:

> There are some quite specialised cases where the advice in the PEP is wrong - it's the same reason we have both "*_Check()" and "*_CheckExact()" type checking APIs at the C level.
> 
> Cases where it's mandatory to use an identity check:
> * when you're making very stringent assumptions about the type's behaviour which may not be valid for subclasses (common in serialisers, also something that often applies for optimisation hacks)
> * when you're needing to identify strict subclasses (i.e. you want to *exclude* the type itself, as we do for operand precedence handling)
> * when you're terminating an otherwise infinite regress (e.g. in a __new__ implementation)
> 
> Cases where it doesn't really matter that much:
> * when the type doesn't allow subclasses
> * where they're technically allowed, but it's an internal class nobody every actually messes with
> 
> For pep8-the-tool, I think it's reasonable to complain by default and require # noqa to say "I know what I'm doing". 
> 
> Cheers,
> Nick.

sigmavirus24:
So I don't think allowing `# noqa` was the right move here. This seems very likely to be a bug in the implementation of the check. It looks like it is checking for `type*` compared to `type*`. I'll see if I can figure this out. @IanLee1521 will you revert this allowance if in fact I can narrow down and fix the bug?

IanLee1521:
Sure, I'd consider it. Though after reflecting on it more, I feel as though giving the control to the end user makes reasonable sense. E.g. (as @ncoghlan said)

> complain by default and require # noqa to say "I know what I'm doing"

sigmavirus24:
@IanLee1521 that's the point on which I disagree with you and @ncoghlan strongly. pep8 more often than not being used as a compliance tool to ensure entire projects are maintaining a defined style. The more checks we enable `noqa` and `nopep8` for, the weaker the compliance tool becomes and the more controversial code reviews become on those code bases. In this case, the check seems to be pretty clearly buggy. As @hongqn and you pointed out, there are several ways to write these checks such they pass but they only fail when comparing to something that starts with "type". That's a pattern that shows there's a bug in pep8.py. Allowing more and more checks to be disabled by `noqa` is simply doing a disservice to the users.

ncoghlan:
The amount of specialised and pre-pep8 code we have in the standard library
is one of the main reasons we *don't* enforce a mandatory style checker
other than reindent.py.

However, I also refer you to the start of PEP 8: trust the humans. PEP 8
compliance should be a *default* position, which can be waived for
sufficiently good reasons, not an inflexibly enforced dogma. If a
development team can't have a civil and productive discussion in a code
review as to whether or not a particular situation justifies the use of the
"noqa" flag, then having pep8 be selective in which warnings can be
disabled will only serve to mask the underlying dysfunctional team
dynamics, not help resolve them.

sigmavirus24:
I think we're each focusing too much on separate target audiences. I've seen more than enough contributions to open source projects where the contributor refuses to fix something and just `# noqa`'s it in spite of the requests of the maintainers because they feel they're exerting ownership over those 5 lines they wrote and they want it to be their style. Pro-active maintainers will just merge it and fix it themselves but they shouldn't have to do that just because `# noqa` was used out of sheer laziness.

Beyond the language in the document, the noise created by adding `# noqa` is something @florentx has wanted to limit. By limiting the errors and warnings that can be silenced via `# noqa` to legitimately controversial codes, this has henceforth been effective. You can still find `# noqa` polluting projects but the most I've seen it in a single file thus far has never exceeded 5 or 7 times which seems like more of a victory than anything else.

ncoghlan:
One of the one hand, if someone refuses to fix their patch, maintainers should really be rejecting their pull request, rather than doing the work the prospective contributor refused to do (the primary exception being when the maintainer really wants the relevant change themselves). Getting patches rejected for policy non-compliance is the only way a lot of people are likely to learn that rules exist for a reason and do apply to them, even if the tools allow them to override the warning.

On the other hand, if folks actually want fine-grained control over their linting, pylint is likely to be a better choice in general, so perhaps another possible approach would be to explicitly recommend pylint for cases where folks want that kind of fine-grained control, rather than adding more knobs and dials to pep8?

I think @sigmavirus24 is right that we're talking about two significantly different audiences here, and encouraging folks to migrate away to a different tool that is better suited to their needs can be a good way to prevent scope creep.

sigmavirus24:
pep8 and pylint definitely cover different audiences. pylint has far too many knobs to be easily used or understood by most beginners. It takes most people I've seen use it a while to settle on a reasonable `.pylintrc`. pep8 is much simpler which I think most people consider a feature. Granted people who come from pylint are going to want to silence every possible error that could be raised but I don't think that's the approach this project should take.

Regardless, I just looked into the test suite for E721 and it looks as if the root cause of this issue was intentional. I'm not sure what the value is in restricting the comparison to something from the `types` module. It seems like a fairly legitimate use case. At this point, I would do some archaeology but it's awfully late here and I've been operating on way too little sleep lately.

PR #314, opened 2014-08-08
Fixing compound_statement not to be quadratic in # of :s

darioush:
`compound_statement` is quadratic in term of # of colons included in a line
This can be problematic for things like generated dictionaries.
I've made it not to be quadratic anymore.


arthall:
Working  on this one.

IanLee1521:
Merged as part of #522

IanLee1521:
Thanks @darioush !

PR #345, opened 2014-11-13
Fix non-ascii characters continued indentation

dongweiming:
I have 2 files for this:

```python
$cat ascii.py
story_tags = [
    (('xxxxxx', 38), [('xxxx', 39),
                      ('xxxx', 40)]),
]
```
and

```python
$cat non-ascii.py
story_tags = [
    (('数码产品', 38), [('手机', 39),
                       ('配件', 40)])
]
```

They look the same format, but the result was different:

```python
$pep8 ascii.py

$pep8 non-ascii.py
non-ascii.py:3:25: E128 continuation line under-indented for visual indent
```
Because:

```python
In [1]: len('xxxx')
Out[1]: 4

In [2]: len('数码')
Out[2]: 6
```

They look same width. But the length is different

### DIFF
--- a/pep8.py
+++ b/pep8.py
@@ -122,6 +122,50 @@ HUNK_REGEX = re.compile(r'^@@ -\d+(?:,\d+)? \+(\d+)(?:,(\d+))? @@.*$')
 # a comment which is on a line by itself.
 COMMENT_WITH_NL = tokenize.generate_tokens(['#\n'].pop).send(None)[1] == '#\n'
 
+# GENERATED DATA
+# generated from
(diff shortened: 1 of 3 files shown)


IanLee1521:
I'm going to go ahead and close this as won't fix due to the [recommendation from PEP-8](https://www.python.org/dev/peps/pep-0008/#source-file-encoding) that string literals and comments should be ASCII encoded (a requirement in Python 3 per [PEP-3131](https://www.python.org/dev/peps/pep-3131)). Specifically: 

> In addition, string literals and comments must also be in ASCII.

PR #363, opened 2014-12-27
sys.stdout.flush() to avoid broken line.

methane:
See https://gitlab.com/pycqa/flake8/issues/17

### DIFF
--- a/pep8.py
+++ b/pep8.py
@@ -1681,6 +1681,13 @@ class StandardReport(BaseReport):
                 'row': self.line_offset + line_number, 'col': offset + 1,
                 'code': code, 'text': text,
             })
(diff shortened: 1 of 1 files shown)


IanLee1521:
So, I tested moving the call to ``sys.stdout.flush()`` as a single call at the end of the for loop and I was still able to verify the fix using the steps outlined at https://gitlab.com/pycqa/flake8/issues/17.

@methane or @sigmavirus24 , could either of you confirm the same results? Thanks.

IanLee1521:
I've added a test support method (73b1e63) and updated the changelog (4579585) as well.

IanLee1521:
Merged in commit bcd5d2b. Thanks for the work on the patch!

Issue #368, opened 2015-01-08
defined config overwritten by config further up directory tree


jsnklln:
In the read_config function first the "user" config file is read then the directory tree is searched for "local" config files.  If any config files exist within the directory tree they overwrite what was defined in the file specified by using --config.   I've made a fork and plan on issuing a pull request soon.  BTW I have no idea what I'm doing.

IanLee1521:
Closed with commit 19d9a82

sigmavirus24:
It seems like the fix for this issue essentially broke people relying on config files in `$HOME/.config/`. See also: https://gitlab.com/pycqa/flake8/issues/33

jsnklln:
I've got to leave for the weekend but I wanted to include what I've found so far:

    flake8 --version
    2.3.0 (pep8: 1.6.1, mccabe: 0.3, pyflakes: 0.8.1) CPython 2.7.9 on Darwin

My python code:
`x = "fsafdsafdsafdsafdsafdsafdsafdsafdsafdsafdsajfkdsajkfdsajkfdsjaklfdjsaklfjdsaklfjdkslajfkldsajklfdsajklfjdsaklfdsa"`

My config:

    [flake8]
    max-line-length=100

Run it:

    flake8 code.py
    code.py:1:101: E501 line too long (119 > 100 characters)

That's goodish news.

In the process I ran across the fact that the .config/pep8 file is read when flake8 is run but the [flake8] section is used.  When pep8 is run the .config/pep8 file is read but the [pep8] section is used.  I haven't looked into why flake8 isn't pulling in the right file.

sigmavirus24:
@IanLee1521 should I file a separate issue for the regression or should we reopen this to cover it?

IanLee1521:
No, I think this was a regression introduced with some cleanup I did. See #380 for more details.

I've tested against flake8 2.3.0 and commit 3bbaa23 and this seems to be resolved.

```
$ cat bar.py
x = "fsafdsafdsafdsafdsafdsafdsafdsafdsafdsafdsajfkdsajkfdsajkfdsjaklfdjsaklfjdsaklfjdkslajfkldsajklfdsajklfjdsaklfdsa"

$ cat ~/.config/flake8
[flake8]
max-line-length=100

$ flake8 --version
2.3.0 (pep8: 1.6.1, mccabe: 0.3, pyflakes: 0.8.1) CPython 2.7.6 on Linux

$ flake8 bar.py
[...]

PR #369, opened 2015-01-08
#368 don't search the direcotry tree if config was defined with --config

jsnklln:
Keeps configs up the directory tree from overwriting what is in the file defined with --config.  (I hope I'm doing this right)

### DIFF
--- a/pep8.py
+++ b/pep8.py
@@ -1922,15 +1922,16 @@ def read_config(options, args, arglist, parser):
             print('user configuration: %s' % user_conf)
         config.read(user_conf)
 
-    local_dir = os.curdir
-    parent = tail = args and os.path.abspath(os.path.commonprefix(args))
-    while tail:
-        if config.read([os.path.join(parent, fn) for fn in PROJECT_CONFIG]):
-            local_dir = parent
-            if options.verbose:
-                print('local configuration: in %s' % parent)
-            break
-        (parent, tail) = os.path.split(parent)
+    else:
+        local_dir = os.curdir
+        parent = tail = args and os.path.abspath(os.path.commonprefix(args))
+        while tail:
+            if config.read([os.path.join(parent, fn) for fn in PROJECT_CONFIG]):
+                local_dir = parent
+                if options.verbose:
+                    print('local configuration: in %s' % parent)
+                break
+            (parent, tail) = os.path.split(parent)
 
     pep8_section = parser.prog
     if config.has_section(pep8_section):


IanLee1521:
Hi @jsnklln -- Thanks for the PR!

After looking into this a bit more, I think perhaps a little more care is needed. Your solution only allows for the config specified on the command line (defaulting to: ``$HOME/.config/pep8``) to be read if it exists (if it doesn't it will check the directories above it). Instead I think something like the following order should be implemented:

1. Read ``$HOME/.config/pep8``
2. Read config files from higher directory levels most likely in reverse order (``/`` -> ``current directory``)
3. Read config file specified on commandline (``--config foo.ini``)

In this case each successive level will overwrite any values read above (see ``ConfigParser.ConfigParser.read()``). 

In any event the documentation can definitely be cleaned up to make this all clearer, both in ``read_config()`` and in the ``pep8 -h`` section on "Configuration".

jsnklln:
That would mean that configs defined in `$HOME/.config/pep8` will exist even when they're not defined in the file given with `--config`.  I don't think that's what most users expect, at least it's not what I would expect.  When I explicitly set the config file I expect my configs to be explicitly set. I see where you're going with this and I don't hate the idea I just think it's confusing to users.  Something like this is already accomplished by changing the config file in the current directory.  

If it helps my changes should, although they probably don't, throw an error if you define a config that doesn't exist.  (At least that's what I would expect.)

IanLee1521:
Ah, I see where the confusion is. No, they won't actually have to exist. ``ConfigParser.ConfigParser.read()`` handles missing files automatically by just not reading them in. So you can safely read in from ``$HOME/.config/pep8`` and if the file doesn't exist, that's handled by ConfigParser.

Also note that with the read order I specified in my last post, since the ``--config`` specified file is handled (read) last it will overwrite any values already been read in (see below for an example).

Example of how this works (note, this is all about config parser behavior rather than pep8 behavior):

```python
In [1]: import os, ConfigParser

In [2]: c = ConfigParser.RawConfigParser()

In [3]: os.path.exists('/home/ianlee1521/.config/pep8')
Out[3]: False

In [4]: cat foo.ini
[pep8]
select = E,W

In [5]: cat bar.ini
[pep8]
select = E

In [6]: c.read('/home/ianlee1521/.config/pep8')
Out[6]: []

In [7]: c.read('foo.ini')
Out[7]: ['foo.ini']

In [8]: c.get('pep8', 'select')
Out[8]: 'E,W'

In [9]: c.read('bar.ini')
Out[9]: ['bar.ini']

In [10]: c.get('pep8', 'select')
Out[10]: 'E'
```

jsnklln:
I'm more worried about values that would be set by reading `$HOME/.config/pep8` but aren't defined in the file specified by `--config`.  

For example:
$HOME/.config/pep8 is:
[pep8]
max-line-length=1000

myconfig.ini is:
[pep8]
ignore=E226

So I run `pep8 --config myconfig.ini`, I'm expecting it to balk on lines with a length greater than 79 (the default per the docs) but it's not.  Now I'm completely confused because I'm sure I'm not setting max-line-length in the config I specified and it's late and dinner's already on the table and I've got to take the kids to karate practice and......well you get the idea.

The inverse is also true.  I set `max-line-length = 20` in .config/pep8 and 6 months later can't figure out why pep8 is complaining about *every single line*, when I use `--config` of course.

sigmavirus24:
So I'm personally a bit conflicted by this. On the one hand, I agree with @jsnklln that this is the behaviour most users expect. On the other hand, we can't change this without potentially breaking this for *someone*. Someone, somewhere, is unknowingly depending on their user config being loaded and integrated with their per-project one that they specify using `--config`. Regardless, I think doing this now is the right thing. I haven't tested @jsnklln or really reviewed the code so that's still up for debate, but the functionality as @jsnklln proposes is (IMO) correct.

As a side note @IanLee1521, the proper way to search for & merge configurations is to go from 

```
/ -> /etc{/project_dir} -> $HOME/ -> XDG-HOME -> .
```

Lots of projects even skip looking for `/.pep8` and go right to `/etc{/project_dir}`.

IanLee1521:
@jsnklln -- I feel like if you are pointing directly at a config, and you want it to be "max-line-length=79" then that should be set in the config you are manually specifying, e.g.:

`$HOME/.config/pep8`:
```ini
[pep8]
ignore=E226
```

`./myconfig.ini`:
```ini
[pep8]
max-line-length=79
ignore=E226
```

The issue seems to be should the command line specified config override all other configs (`/etc`, `$HOME`, etc) even if the option is not explicitly set in the `--config` config file? Or should there be a merging up the stack as @sigmavirus24 mentions, and only overriding when explicitly desired? Think of it as a sequence of `dict.update()` calls.

@sigmavirus24 -- That is the progression I intended to mean. I think my use of the term "project config" was misspoken. I was thinking of the `setup.cfg`, `tox.ini`, and `.pep8` directories in `.`.  My reference to the reverse order point:

> Read config files from higher directory levels most likely in reverse order (/ -> current directory)

Was in regards to a case where you might have a different `.pep8` config file in both a top level dir of a project, and a different configuration in a sub directory in the same project.  Something like:

* top_dir
  * .pep8
  * sub_dir
    * .pep8
    * foo.py

and then calling pep8 from the sub dir
```bash
cd top_dir/
pep8 sub_dir/foo.py  # currently will use only the `top_dir/.pep8` configuration

cd top_dir/sub_dir/
pep8 foo.py  # currently will use only the `top_dir/sub_dir/.pep8` configuration
```

Note: This might be an unreasonable use case, so maybe it doesn't really matter.

IanLee1521:
I committed a branch (ece1bde) that does what I am describing (e.g. more explicit handling). Currently one of the test cases is failing, this is because it is picking up the `pep8/setup.cfg` which defines `ignore=E226,E24`.

jsnklln:
@IanLee1521 I don't know what else to say except I think that behaviour is just plain weird and unexpected by most if not all users.  Basically I work under the idea that config files, especially when I define them explicitly on the commandline, are modifications to the defaults not modifications to the reading of the default config files.  In an attempt to find some common ground: what about an option that would state what configs are set by what files?  That would at least make debugging anything that seems weird easier.  I'm thinking that when configs are read just print out the filename and contents.  I'm not sure how that would jive with things like flake8.

IanLee1521:
I can see your point. In that case I'm alright with changing the behavior, but I think that the default value of the ``--config`` should be changed to ``None``, currently computed as:

``` python
try:
    if sys.platform == 'win32':
        DEFAULT_CONFIG = os.path.expanduser(r'~\.pep8')
    else:
        DEFAULT_CONFIG = os.path.join(os.getenv('XDG_CONFIG_HOME') or
                                      os.path.expanduser('~/.config'), 'pep8')
except ImportError:
    DEFAULT_CONFIG = None
```

The issue then remains (mostly) as to how to merge the home and local configurations. I'm inclined to keep the behavior of merging those two types of configurations via something like what I described in my post above. Thoughts?

As an aside, I think that having something like a ``--show-config`` option that would print out what files are being read and what config parameters are coming from them would be a good addition, and handle some of the potential issues with the way that the local configuration is read in.

sigmavirus24:
So I agree that `--config` should be the one source of truth when specified. I also agree that `--config`'s default value should be `None`, but it should be documented that without specifying it, it will read configuration files and merge them.

I don't think `--show-config` is an option we should be adding though. That's a debugging flag at best and we already have a way to show more verbose output. Anything you think should be enabled by `--show-config` should instead be printed when using a specific verbosity level. I think, further, that this is already handled in [`read_config`](https://github.com/jcrocholl/pep8/blob/bcd5d2bae39bc2b43c02f394fea9950f40099199/pep8.py#L1934..L1961). It also looks, from reading that function, as if it merges them in the proper order, i.e., the user's configuration is read first, then the project config is read and merged into it.

IanLee1521:
Alright, I have finally gotten the time to finish working on this. I've updated it with the behavior discussed. The only remaining (potential) issue is that in the case that there are multiple directories (between '/' and current dir) that contain local configurations, only those in highest (closest to current dir) will be read.

IanLee1521:
Merged with commit 19d9a82 (this commit was rebased and is now 57372c6.

PR #383, opened 2015-02-19
Fix problem with treating ~ operator as binary.

### LINKED ISSUE #384: Tilde is treated as binary operator
Following code cause W503 to be reported, because ~ is treated as a binary operator, where in reality it is unary.
```
var = (1 &
       ~2)
```

czyzykowski:
Fixes #384

### DIFF
--- a/pep8.py
+++ b/pep8.py
@@ -1017,13 +1017,14 @@ def break_around_binary_operator(logical_line, tokens):
     Okay: x = '''\n''' + ''
     Okay: foo(x,\n    -y)
     Okay: foo(x,  # comment\n    -y)
+    Okay: var = (1 &\n       ~2)
     """
     def is_binary_operator(token_type, text):
         # The % character is strictly speaking a binary operator, but the
         # common usage seems to be to put it next to the format parameters,
         # after a line break.
         return ((token_type == tokenize.OP or text in ['and', 'or']) and
-                text not in "()[]{},:.;@=%")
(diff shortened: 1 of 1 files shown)


mbarrien:
Ping on getting this merged? Affecting my projects.

Incognito:
Any chance this can get merged? @IanLee1521 @Nurdok @PCManticore @sigmavirus24 (@-ing people public on the PyCQA org).

sigmavirus24:
@Incognito the only people who are allowed to merge pull requests here are @IanLee1521 and myself. Please don't try to @-mention the entire organization.

As for this pull request, the original issue has people claiming this is an incomplete fix. If you want to make sure this is fixed, could you verify that concern? If this works for your case and the other cases described there, I'll be happy to merge it if @IanLee1521 can't get around to it.

sigmavirus24:
IMO, the person saying this didn't fix the case with `-` seems to be incorrect. The docstring tests clearly test that case so I think that case is already covered. This also looks to fix the original bug. I'll try to test this tonight.

IanLee1521:
So I looked at it, and it there is a warning if you add a doctest of `Okay: var = (1 &\n       -2)` or `Okay: var = (1 *\n       -2)`, the W503 warning does pop up. I'm going to open a separate issue for that, and merge / close this one which is specific to the `~` operator.

Incognito:
Thanks for getting this through! I appreciate having this fix.

PR #385, opened 2015-02-21
doc warning fix

mindw:


docs/intro.rst
```diff
@@ -237,7 +237,7 @@ This is the current list of error and warning codes:
 | E116     | unexpected indentation (comment)                                     |
 +----------+----------------------------------------------------------------------+
 +----------+----------------------------------------------------------------------+
-| E121 (*^)| continuation line under-indented for hanging indent                  |
+| E121(\*^)| continuation line under-indented for hanging indent                  |
```
sigmavirus24: I think we still want it to be:

```rst
| E121 (\*^) |
```

Thanks for noticing that the `*` wasn't appearing properly though.

docs/conf.py
```diff
@@ -172,7 +172,7 @@
 #html_file_suffix = None
 
 # Output file base name for HTML help builder.
-htmlhelp_basename = 'pep8doc'
+htmlhelp_basename = 'pep8'
```
sigmavirus24: I'm not sure I quite get this change though. =(

docs/intro.rst
mindw: This will require to re-align the entire table.

sigmavirus24: All of the other errors have spaces between the `(` and the error code. Re-aligning the table is quite simple to do if you use vim or sublime text

docs/conf.py
mindw: Use the same file name for CHM output as for EPUB, QtHelp, pdf etc.

mindw:
Well, using sublime + Table Editor did the trick.

docs/intro.rst
```diff
@@ -217,191 +217,191 @@ Error codes
...
-| W601     | .has_key() is deprecated, use 'in'                                   |
-+----------+----------------------------------------------------------------------+
-| W602     | deprecated form of raising exception                                 |
-+----------+----------------------------------------------------------------------+
-| W603     | '<>' is deprecated, use '!='                                         |
-+----------+----------------------------------------------------------------------+
-| W604     | backticks are deprecated, use 'repr()'                               |
-+----------+----------------------------------------------------------------------+
++------------+----------------------------------------------------------------------+
+|    code    |                            sample message                            |
```
sigmavirus24: Why did you center "code" and "sample message"?

docs/conf.py
sigmavirus24: Sure, but it's unrelated to fixing the formatting of the table and it doesn't have to be the same name.

mindw:
sublimetext 3 + Table Editor = center  "code" and "sample message".

Does it matter?

docs/conf.py
mindw: it saves me from keeping a private branch with the fix. The `doc` suffix seems like a bad default anyhow.

sigmavirus24:
> Does it matter?

I'm not sure it's a blocker, but it's confusing to see as part of this change. It's also unclear if docutils/sphinx will render this differently as a result.

mindw:
Let me rephrase:

1. No, I did not center the headings intentionally.
2. I could not notice and change in the output:
Original - http://rst.ninjs.org/?n=67d2461d8e0e40a9eb3bb83a13593cc1&theme=basic
Patch - http://rst.ninjs.org/?n=06979870b8cf9567972c18e8c9c0694d&theme=basic
3. Given the above do you want this change reverted? kept?

IanLee1521:
This appears to have been fixed in 31c0215a03264c8c3f4ae8bd59d46c091b03451f and 37648d3fb74db1fd952c7cb101eaf7224055ce1b, thanks for the report @mindw !

PR #395, opened 2015-03-21
Python2 now detects encoding according to PEP263

### LINKED ISSUE #391: Variable-lenght encoding (like UTF-8) confuses E128
E128 gets confused with variable-lenght (with multi-byte) encoding, such as UTF-8. Using a multi-byte character makes PEP8 return an E128 warning of under-indent. Tests show that it expects one more space for each multi-byte character, consistent with UTF-8 double-bytes for the extended I'm using.

Apparently PEP8 is either ignoring the shebang line or using len() directly with UTF-8 string without converting to unicode first (even with the 'u' prefix).

Example `pep8utf8.py`:
```python
# -*- coding: utf-8 -*-
_EPSILON = {'pressao': {'psig': 1,
                        'kPag': 5},  # No error here
            'pressão': {'psig': 1,
                        'kPag': 5},  # Error
            }

```
Running `pep8 --show-source pep8utf8.py` yields:
```
pep8utf8.py:5:25: E128 continuation line under-indented for visual indent
                        'kPag': 5},  # Error
                        ^
```
**Note 1:** I have tried using unicode strings, but the error is still there.
**Note 2:** Python 2.7.6 |Anaconda 2.1.0 (32-bit)| (default, Nov 11 2013, 10:50:31) [MSC v.1500 32 bit (Intel)] on win32

miszobi:
This adds support for detecting source file encoding as defined in PEP263 in Python2.
- inlined python's 3 codecs.detect_encoding to make python2 recognize
  PEP 263 encodings the same way as the python3 implementation does
- simplified readlines implementation a little bit, to share the code
  between 2 and 3
- added the test case mentioned in the issue, and a shift-jis variant
- build passing at https://travis-ci.org/miszobi/pep8/builds/55316861
- fixes #391

### DIFF
--- a/pep8.py
+++ b/pep8.py
@@ -55,12 +55,11 @@ import inspect
 import keyword
 import tokenize
 import codecs
-import io
-from io import TextIOWrapper
 from optparse import OptionParser
 from fnmatch import fnmatch
 try:
     from configparser import RawConfigParser
+    from io import TextIOWrapper
 except ImportError:
     from ConfigParser import RawConfigParser
 
@@ -1199,8 +1198,8 @@ def detect_encoding(readline):
     bom_found = False
     encoding = None
     default = 'utf-8'
-    BOM_UTF8 = b'\xef\xbb\xbf'
-    blank_re = re.compile(br'^[ \t\f]*(?:[#\r\n]|$)')
+    BOM_UTF8 = '\xef\xbb\xbf'
+    blank_re = re.compile('^[ \\t\\f]*(?:[#\\r\\n]|$)')
 
     def get_normal_name(orig_enc):
         """Imitates get_normal_name in tokenizer.c."""
@@ -1217,7 +1216,7 @@ def detect_encoding(readline):
         try:
             return readline()
         except StopIteration:
-            return b''
+            return ''
 
     def find_cookie(line):
         cookie_re = re.compile(r'^[ \t\f]*#.*coding[:=][ \t]*([-\w.]+)')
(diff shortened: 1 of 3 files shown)


pep8.py
```diff
@@ -54,11 +54,13 @@
 import inspect
 import keyword
 import tokenize
+import codecs
+import io
```
miszobi: Just noticed this isn't available in Python 2.5, which seems to still be supported (based on [the changelog](https://pypi.python.org/pypi/pep8)). 

Is this still the case? I'll rework if so.

myint:
By the way, I'm not sure if it is relevant, but in `autopep8`, I just [imported](https://github.com/hhatto/autopep8/blob/bb81d1922970c485edd5dd75f279120c709dc1fe/autopep8.py#L139) the code rather than copying it.

pep8.py
sigmavirus24: I believe 2.5 support is still planned but @florentx will have to confirm that.

miszobi:
@myint thanks for the tip, didn't know you could do that. Unfortunately it seems that lib2to3 is also unavailable in python2.5, so I can't use that if we still aim to support it.

valgur:
This would be a useful improvement for pep8, too bad it has not been already merged.

One thing that should be noted though, I think, is that this pull request does not fix the #391 issue entirely: it does nothing when the input is provided via stdin ('-' as the file name) or the `lines` parameter in `StyleGuide.input_file()` and `Checker()`.

hayd:
**Is** python 2.5 supported?

myint:
Good question. Also, the last CPython 2.5 release was in [2011](https://www.python.org/download/releases/2.5.6/).

dusek:
Isn't Python 2.7 the only sensible 2.x release to support now? Or is anything else blocking this? Just hit the bug #391, so wondering.

asottile:
pycodestyle is py3+ now so this isn't relevant

PR #396, opened 2015-03-26
Allow "empty" file to have a \n character

gpoulin:
W292 and W391 are conflicting when it comes to "empty" files. Normally empty files are not that useful. However in Python, it's often useful to have an empty `__init__.py` to declare a folder as a package without necessarily putting any logic there (In a perfect world those file would contain a docstring).

We could debate whether or not an "empty" file should contain a `\n` or not. However, I don't see any real practical implication and think pep8 should simply accept both.

### DIFF
--- a/pep8.py
+++ b/pep8.py
@@ -189,7 +189,7 @@ def trailing_blank_lines(physical_line, lines, line_number, total_lines):
     """
     if line_number == total_lines:
         stripped_last_line = physical_line.rstrip()
-        if not stripped_last_line:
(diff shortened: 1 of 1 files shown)


gpoulin:
I wanted to add unit test for that, but I wasn't sure how to test this case using the testsuite.

sigmavirus24:
@gpoulin best scenario may be to just add a comment explaining that behaviour. I have ideas for better testing for pep8 in the future but the future is not now ;)

Issue #45, opened 2011-12-29
Indentation checkers should not be active on code inside triple-quotes


posborne:
If I have a multi-line string (in triple quotes), I don't believe that pep8 should be analyzing the contents for indentation problems and the like.  This code may very well be not-python and, as such, need not comply to pep8.  It might be nice (though I am not certain of how easy) to treat docstrings different from other block quotes as there are guidelines on docstrings in PEP 257: http://www.python.org/dev/peps/pep-0257/

maxkoryukov:
@florentx @myint 

> `# noqa` on the last line of the docstring.

Is it workaround, or final solution? I would like to use docstrings for documentation purposes, and I want to have whitespaces, tabulation and other stuff there, out of the box, without sane limitations. Currently it is looks like `pep8` poke its nose in where it didn't belong (IMO).

For example, I want to place markdown into docstrings, bcz doc-parser supports it, and I like it. But `flake8` raise an issue on my docstrings, because there are trailing whitespaces (_they are required in my cases_)

BTW: PyCharm **can not** remove trailing whitespaces in docstrings automagically. It requires some tunings/dancing/virgin-blood/moonlight to provide this functionality. Sad, because people told, that PyCharm - is a _decent_ IDE for python..

sigmavirus24:
> Is it workaround, or final solution?

It is something you can use _today_. So in a way it's both.

> Currently it is looks like pep8 poke its nose in where it didn't belong (IMO).

I disagree, but that's *my* opinion.

> there are trailing whitespaces (they are required in my cases)

That's unfortunate

> BTW: PyCharm can not remove trailing whitespaces in docstrings automagically.

That's surprising and disappointing but not relevant here.

maxkoryukov:
@sigmavirus24 thank you for reply. I will cry and eat this cactus ))

gyermolenko:
>> Is it workaround, or final solution?

> It is something you can use today. So in a way it's both.

@sigmavirus24 Could you please clarify resolution on this issue?
Personally I can't understand why it is closed (just my opinion, no hard feelings)

sigmavirus24:
> > > Is it workaround, or final solution?
> 
> > It is something you can use today. So in a way it's both.
> 
>  Could you please clarify resolution on this issue?

- `# noqa` will continue to skip this warning. (hence the joke about finality)
- `# noqa` support was added providing an escape hatch for folks who want to opt out of this behaviour
[...]

PR #397, opened 2015-03-29
Allow spaces after tabs for alignment

dsedivec:
First off, thanks for writing pep8!  It's very useful.

We've got a large code base that uses all tabs.  I'm a bit pedantic when it comes to indentation, so I like to make sure that my hanging indents line up at any tab width by aligning them with spaces after the tabs.  For example:

    if True:
        long_name(foo, bar,  # imagine this is indented with one tab
                  baz)       # this is indented with one tab plus ten spaces

Mixing spaces and tabs like this means that, despite using the hated tabs for indentation, you get properly lined-up code no matter what width people use for displaying tabs.

Python has no problem with this since it's mixed spaces and tabs inside parentheses.  pep8, though, wasn't able to make that distinction with its physical line check.  My commit thus replaces the physical check for E101, mixed tabs and spaces, with a logical line check.  Using a logical line check made it easier to examine the context, to know when mixed tabs and spaces should be permitted.

PEP 8 says, "Hanging indents \*may\* be indented to other than 4 spaces." However pep8's code and tests don't seem to permit this, so my replacement check doesn't permit it either.  For example, pep8 still gives "E101 indentation contains mixed tabs and spaces" on the last line here:

    if True:
        foo()   # imagine these two lines are indented with
        bar(    #   a single tab
          baz)  # but this line is one tab plus two spaces

One test was moved from E10.py to W19.py, since the test correctly no longer produces E101.  Two tests in W19.py have had E101 removed.  The first was using spaces after tabs in a permitted way, for alignment.  The second had mixed tabs and spaces in a multi-line string, and I saw no reason that should be forbidden (what you do in strings is your business).  Finally, one test was added to each of E10.py and W19.py to catch some tricky cases in my own code.

I'm happy to make further explanations, add tests or comments, etc. as required.

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -56,6 +56,8 @@ import inspect
 import keyword
 import tokenize
 import warnings
+import itertools
+import operator
 from optparse import OptionParser
 from fnmatch import fnmatch
 try:
@@ -96,6 +98,7 @@ WS_OPTIONAL_OPERATORS = ARITHMETIC_OP.union(['^', '&', '|', '<<', '>>', '%'])
 WS_NEEDED_OPERATORS = frozenset([
     '**=', '*=', '/=', '//=', '+=', '-=', '!=', '<>', '<', '>',
     '%=', '^=', '&=', '|=', '==', '<=', '>=', '<<=', '>>=', '='])
+CLOSING_OPERATORS = {"(": ")", "{": "}", "[": "]"}
 WHITESPACE = frozenset(' \t')
 NEWLINE = frozenset([tokenize.NL, tokenize.NEWLINE])
 SKIP_TOKENS = NEWLINE.union([tokenize.INDENT, tokenize.DEDENT])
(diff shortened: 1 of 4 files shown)


myint:
I guess making this a logical check may also fix #45 as a side benefit.

It seems kind of strange that Travis CI never works with pull requests for the pep8 repository specifically.

gg7:
+1

sigmavirus24:
@gg7 do you mean to say that you'll be working on fixing this for us? That'd be awesome.

gg7:
@sigmavirus24 there's a patch available. Have you tried it?

dsedivec:
Not quite sure what to make of @sigmavirus24's comment!  If there's more work or explanation that needs work on this, please let me know.

IanLee1521:
I believe this behavior should not be allowed as part of the check due to the issues that PEP-8 has with tabs in general and mixing spaces and tabs in particular. Specifically:

> Python 3 disallows mixing the use of tabs and spaces for indentation.
> 
> Python 2 code indented with a mixture of tabs and spaces should be converted to using spaces exclusively.

-- Source: https://www.python.org/dev/peps/pep-0008/#tabs-or-spaces

hyperair:
@IanLee1521, 

PEP8 disallows *indenting* using a mixture of tabs and spaces, but does not say anything about alignment.

The paragraph you quoted also follows up with:
> When invoking the Python 2 command line interpreter with the -t option, it issues warnings about code that illegally mixes tabs and spaces. When using -tt these warnings become errors. These options are highly recommended!

This [sample code](https://gist.github.com/hyperair/9a70d88749a8782f2902771942b06074), demonstrating tabs for indentation and spaces for alignment is accepted by both Python 3 and Python 2 with `-tt`. Ergo, it isn't illegal.

Please consider reopening this PR.

alexander255:
Will this be reconsidered?

sigmavirus24:
Most likely not

alexander255:
@sigmavirus24: Well @hyperair's argument still stands, but I know you don't care either way. Have a nice day…

Issue #400, opened 2015-04-07
Require two blank lines *after* a def/class too


gvanrossum:
The pep8 tool insist on two blank lines before a top-level class or def, but not *after* one. So you get the following asymmetry:
```
import stuff


def main():
    blah, blah

if __name__ == '__main__':
    main()
```
The intention of PEP 8 is to require two blank lines before the 'if' too. I realize the PEP isn't fully clear, since it only says "separate [...] with two blank lines", and if you want an authoritative rule I can rewrite it. Just let me know!

gvanrossum:
Oh, and clarification: it isn't the 'if' that makes the requirement for two lines. It is the end of the 'def'. The rule I meant to describe is that top-level class and def blocks must be *surrounded* by two blank lines. (But we need to have an allowance for comments too -- I'll let you figure that out by looking at actual code.)

IanLee1521:
I believe, though I'm not able to verify this right now, that comments are ignored if before top level definitions. I will double check the exact rules though.

As far as a rule goes, I'm concerned with files such as (specifically lines 98 - 120):

https://hg.python.org/cpython/file/a48e76252952/Lib/json/__init__.py

Where there are __*__ assignments and imports that are not two newline separated. Should the two line rule apply only to between top level "class ...", "def ...", and "if __name__ ..."  definitions?

gvanrossum:
Only to top-level class and def.

jayvdb:
There is ambiguity in the PEP, and differing opinions naturally, regarding whether 'top-level' refers to indentation or scope. e.g. should two blank lines be enforced when the `def` is within an `if`?
See #366 and #168 for discussions about that.

gvanrossum:
By top level I meant scope-wise, not I dentation-wise.

--Guido (mobile)

jayvdb:
Thank you for the clarification.

When "top-level" is scope wise, there is one area where it becomes a little ugly IMO ... between the `if` and a classdef/def, there should be two lines? e.g.

```python
if True:


    class Foo: pass


...
```

gvanrossum:
Yes, go with what feels right to you there.

--Guido (mobile)

sigmavirus24:
By defining top-level as scope-wise, won't this also mean something like:

```py
def func0(foo, bar, bogus):
    biz = ...
    baz = ...


    def inner_func(other_param):
        # ...


    return inner
```

sigmavirus24:
[...]

PR #402, opened 2015-05-20
Require two blank lines after toplevel def/class; issue #400

benesch:
Here's my attempt to fix issue #400. Not sure what the backwards compatibility policy on this'll be, so I left it as enabled by default.

### DIFF
--- a/pep8.py
+++ b/pep8.py
@@ -234,7 +234,8 @@ def maximum_line_length(physical_line, max_line_length, multiline):
 
 
 def blank_lines(logical_line, blank_lines, indent_level, line_number,
-                blank_before, previous_logical, previous_indent_level):
+                blank_before, previous_logical, previous_logical_toplevel,
+                previous_indent_level):
     r"""Separate top-level function and class definitions with two blank lines.
 
     Method definitions inside a class are separated by a single blank line.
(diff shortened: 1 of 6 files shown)


pep8.py
```diff
@@ -268,6 +270,9 @@ def blank_lines(logical_line, blank_lines, indent_level, line_number,
                 yield 0, "E301 expected 1 blank line, found 0"
         elif blank_before != 2:
             yield 0, "E302 expected 2 blank lines, found %d" % blank_before
+    elif (logical_line and not indent_level and blank_before != 2 and
+          previous_logical_toplevel.startswith(('def', 'class'))):
+        yield 0, "E305 expected 2 blank lines before, found %d" % blank_before
```
xZise: Shouldn't that expect 2 blank lines **after** the `def` or `class`? Sure it's checking for the next logical line if before it there are two lines but only because the block above was a `def` or `class`.

Maybe also mention that it expects two lines after a method-level function or class (I know E302 doesn't mention that too but that may be improved too?).

benesch: Hmm? I'm pretty sure the condition is right: it's looking for a non-indented line preceded by more or less than two blank lines where the preceding non-blank non-indented line was a function or class definition.

It doesn't actually expect two lines after a method-level function or class, since those are only supposed to have one separating blank line according to PEP 8.

xZise: The condition is right but I find the message confusing.

benesch: Ah, agreed! I'll defer that to another diff, though.

ArcTanSusan:
This hasn't been worked on since August 2015 last year. For PyCon Sprints 2016, I've fixed the conflicts in this new branch and rebase-squashed both commits into 1 commit: https://github.com/PyCQA/pycodestyle/pull/536

PR #413, opened 2015-07-03
remove EOL with backslash

dongweiming:
I found a flake8 problem. When i use exclude with a long long excluded files. i need make the long commands split over multiple lines. like this:

```python
flake8 --ignore=E501,F403,E265,F812 --exclude='test1.py, \
test2.py' 
```

options.exclude is ['test1.py', '\\ntest2.py']

It's not correct

[flake8#31](https://gitlab.com/pycqa/flake8/merge_requests/31). @sigmavirus24 think should be here to solve

### DIFF
--- a/pep8.py
+++ b/pep8.py
@@ -1282,7 +1282,7 @@ def normalize_paths(value, parent=os.curdir):
         return value
     paths = []
     for path in value.split(','):
-        path = path.strip()
+        path = path.strip().strip('\\\r\n')
         if '/' in path:
             path = os.path.abspath(os.path.join(parent, path))
         paths.append(path.rstrip('/'))


dongweiming:
In flake9#33 i just show a very short case. In fact, I want to excluded files number is larger than 30.

sigmavirus24:
I was a bit curious about this, and now I'm wondering why you've chosen to use single quotes for your argument. Take for example this script:

```
# optstest.py
import optparse

parser = optparse.OptionParser()
parser.add_option('--exclude', dest='exclude')

(options, args) = parser.parse_args()

print('Args: %s' % args)
print('options.exclude: %s' % options.exclude)
```

And these executions of it:

```
~ ❯❯❯ python optstest.py --exclude=foo.py,\
bar.py,\
biz.py
Args: []
options.exclude: foo.py,bar.py,biz.py
~ ❯❯❯ python optstest.py --exclude='foo.py,\
bar.py,\
biz.py'
Args: []
options.exclude: foo.py,\
bar.py,\
biz.py
```

Obviously the second exection doesn't display that `\n` but it's there.

I'm not saying that quoting the argument isn't valid in some cases, but this can just as easily be solved by not quoting it. Is there a specific reason you need to quote the values you pass to `--exclude`?

dongweiming:
@sigmavirus24 I think I can explain in more detail. This is a big project. there are a lots of code written by    programmers that has left. and some services code autogenerated by Thrift. It takes a lot of effort to maintain. so i hope i can ignore this part.

dongweiming:
oh i understand

sigmavirus24:
That did not answer my question. Why are you using quotes for the exclude argument?
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

sigmavirus24:
Email is so slow ignore me

PR #420, opened 2015-07-14
Support Python 3.5

myint:
This involves fixing a test and avoiding a function deprecated since Python 3.0.

https://docs.python.org/dev/library/inspect.html#inspect.getargspec

### DIFF
--- a/pep8.py
+++ b/pep8.py
@@ -1328,7 +1328,7 @@ def register_check(check, codes=None):
                 codes = ERRORCODE_REGEX.findall(check.__doc__ or '')
             _add_check(check, args[0], codes, args)
     elif inspect.isclass(check):
-        if sys.version_info[0] >= 3:
+        if sys.version_info >= (3, 3):
             parameters = list(inspect.signature(check.__init__).parameters)
         else:
             parameters = inspect.getargspec(check.__init__)[0]
--- a/pep8.py
+++ b/pep8.py
@@ -1314,6 +1314,13 @@ if COMMENT_WITH_NL:
 _checks = {'physical_line': {}, 'logical_line': {}, 'tree': {}}
 
 
+def _get_parameters(function):
+    if sys.version_info >= (3, 3):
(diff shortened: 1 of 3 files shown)


myint:
By the way, I think it would be a good idea to enable Travis on pull requests for this repository. Currently pull requests are not tested for some reason. I think it is a Travis CI setting.

sigmavirus24:
I toggled Travis through its control pane (off then on) to re-enable this. Good catch @myint

sigmavirus24:
Can you also add the `- nightly` to this pull request?

sigmavirus24:
`inspect.signature` does not exist on Python 3.2. It looks as though the [deprecation notice for Python 3.0](https://docs.python.org/3.2/library/inspect.html#inspect.getargspec) says to use `inspect.getfullargspec` instead of `inspect.signature`. Even so, we can use signature (it seems) on 3.3 and beyond.

pep8.py
```diff
@@ -1328,7 +1328,12 @@ def _add_check(check, kind, codes, args):
                 codes = ERRORCODE_REGEX.findall(check.__doc__ or '')
             _add_check(check, args[0], codes, args)
     elif inspect.isclass(check):
-        if inspect.getargspec(check.__init__)[0][:2] == ['self', 'tree']:
+        if sys.version_info[0] >= 3:
```
sigmavirus24:
```py
if sys.version_info >= (3, 3):
```

That will suffice for what we need. `inspect.signature` was introduced in 3.3 and not before.

myint:
Should be fixed now.

sigmavirus24:
This looks good to me. :+1:

IanLee1521:
Merged! Thanks!

SylvainDe:
Thanks for fixing this issue! :+1: 
Out of curiosity, does anyone know if this is going to be part of the next pep8 release ? If so, is there any expected release date ?

myint:
I would be interested in a PyPI release as well.

PR #421, opened 2015-07-14
Remove pypy* from allowed failures

sigmavirus24:
The pypy jobs have not failed in a long time. Keeping them in
allowed_failures means that if they start failing, no one will notice
them.

We also start using "sudo: false" since we are not installing anything
with apt and this will speed up build times.


IanLee1521:
Great, thanks!

PR #425, opened 2015-07-29
Print 0 when using --count and there are no violations

kampde:
Right now nothing is printed if there are no violations

### DIFF
--- a/pep8.py
+++ b/pep8.py
@@ -1951,7 +1951,7 @@ def get_parser(prog='pep8', version=__version__):
     parser.add_option('--count', action='store_true',
                       help="print total number of errors and warnings "
                            "to standard error and set exit code to 1 if "
-                           "total is not null")
+                           "total is not zero")
     parser.add_option('--max-line-length', type='int', metavar='n',
                       default=MAX_LINE_LENGTH,
                       help="set maximum allowed line length "
@@ -2118,9 +2118,9 @@ def _main():
         report.print_benchmark()
     if options.testsuite and not options.quiet:
         report.print_results()
+    if options.count:
+        sys.stderr.write(str(report.total_errors) + '\n')
     if report.total_errors:
-        if options.count:
-            sys.stderr.write(str(report.total_errors) + '\n')
         sys.exit(1)
 
 if __name__ == '__main__':


sigmavirus24:
So, I've thought about this for a short while. This kind of makes sense since your asking for the count and the help text for the option is ambiguous. (At least, I had always read it as "Print the count [if the count is non-zero] and exit 1 [if the count is non-zero].) So perhaps, this needs to be written less ambiguously. That said, I can imagine people relying on the absence of output in CI tooling right now to signal whether or not there's a problem (regardless of exit code). And this will break those people's expectations. Since this is a **breaking change**, I would rather the project hold off on this until 2.0.0 (if the project decides to merge this at all.)

kampde:
Hmmm I hadn't thought of it as a breaking change, but I kind of agree it is. Also, I think that not outputting anything anything even though I am asking for the count it is rather odd.

Anyways, whatever the project decides about merging (or when to merge) this is ok by me, my CI builds can handle both cases now :)

petsuter:
+1
The current behavior is somewhat confusing e.g. when calling pep8 as an external tool from an IDE.

How about at least printing ``0`` when both ``-v`` and ``--count`` are passed?

IanLee1521:
I rebased the branch to fix the merge conflicts (06fcbbc624aba51a875c7317b59137efd6e20024 and ef44231707e77cca9548aa0f7f74a9e1022fb6ea). This was missed in the 2.0 release. I'm debating releasing it as part of the 2.1 release anyways despite the potential issue with the backwards compatibility... 

Are the any strong objections to that change? Are there users that we know are expecting the use of ``--count`` to print nothing (including to stderr where this goes)?

If there are still concerns, then I will hold off until we are ready for a 3.0 version, which may be coming (relatively) sooner rather than later.

IanLee1521:
Going to close this in favor of the rebased, non-conflicting #543

kampde:
I prefer to wait for the 3.0 release. What's the point on following a version number scheme if you are going to break it?

IanLee1521:
Merged into the https://github.com/PyCQA/pycodestyle/tree/upcoming-3.0 branch for the 3.0 release

Issue #423, opened 2015-07-22
triple-quoted strings should always use double quote characters


jayvdb:
From https://www.python.org/dev/peps/pep-0008/#string-quotes

"For triple-quoted strings, always use double quote characters to be consistent with the docstring convention in PEP 257 ."

sigmavirus24:
This would seem reasonable but would need to be introduced in 1.7.0 (assuming 1.7.0 is planned).

IanLee1521:
Agreed, I think this would be a fine check to add in 1.7.0

SylvainDe:
As a side-note, this is not added in 1.7.0 as the pull request is still waiting for additional information.

SylvainDe:
So, as per the comments on https://github.com/PyCQA/pycodestyle/pull/428 , should we just close the issue and abandon the pull-request ?

asottile:
agreed, given the resolution in #428

PR #428, opened 2015-08-14
Triple-quoted string should use double quote #423

SylvainDe:
From https://www.python.org/dev/peps/pep-0008/#string-quotes

"For triple-quoted strings, always use double quote characters to be
consistent with the docstring convention in PEP 257 ."

Error number is not very good at the moment.


pep8.py
```diff
@@ -231,6 +231,22 @@ def maximum_line_length(physical_line, max_line_length, multiline):
...
+    E999: s = r'''a'''
+    """
+    triple = "'''"
+    for token_type, text, start, end, line in tokens:
+        text = remove_string_prefix(text)
+        if token_type == tokenize.STRING and \
+           text.startswith(triple) and \
+           text.endswith(triple):
+            yield start, "E999 triple-quoted strings "\
+                    "should use double quote characters"
```
IanLee1521: This introduces a pep8 violation that causes the tests to fail.

pep8.py
```diff
@@ -862,8 +887,8 @@ def module_imports_on_top_of_file(
 
     Okay: import os
     Okay: # this is a comment\nimport os
-    Okay: '''this is a module docstring'''\nimport os
-    Okay: r'''this is a module docstring'''\nimport os
+    Okay: ""\"this is a module docstring""\"\nimport os
```
IanLee1521: Why the `\` here and in other lines?

IanLee1521:
Overall I'm happy with this and will merge it once the tests are fixed. I agree we might want a better number for it.

docs/intro.rst
```diff
@@ -372,6 +372,8 @@ This is the current list of error and warning codes:
 +----------+----------------------------------------------------------------------+
 | E902     | IOError                                                              |
 +----------+----------------------------------------------------------------------+
+| E999     | Triple quoted string uses simple quotes                              |
```
sigmavirus24: Why make this an `E9` class error? These are for **Runtime** errors.

pep8.py
```diff
@@ -231,6 +231,22 @@ def maximum_line_length(physical_line, max_line_length, multiline):
...
+    r"""Triple quotes should use double quote characters.
+    Okay: s = ""\"a""\"
+    Okay: s = r""\"a""\"
+    E999: s = '''a'''
+    E999: s = r'''a'''
+    """
+    triple = "'''"
+    for token_type, text, start, end, line in tokens:
+        text = remove_string_prefix(text)
+        if token_type == tokenize.STRING and \
```
sigmavirus24: Please follow [the project convention](https://github.com/SylvainDe/pep8/blob/3761020583a93819c43dca65ccfcfa30d65ce93e/pep8.py#L534..L536). Do not use `\` for continuation.

sigmavirus24:
In general, I'm in favor of this. This will have to be scheduled for 1.7.0 though. It cannot go into a 1.6.x release. Also please address the review left by Ian and me.

Cheers,

pep8.py
SylvainDe: Will do!

pep8.py
SylvainDe: Will do

docs/intro.rst
SylvainDe: I decided to chose a number I'd be able to recognise. Any suggestion is welcome. It doesn't seem to fit in any existing category so I'll leave you guys to decide.

pep8.py
SylvainDe: The trick is to have nested triple quoted strings. Not quite sure what is the best way to do so. Escaping every single double quote character brings a lot of visual noise. Escaping one was enough. Not quite sure what the favourite way to do it is.

sigmavirus24:
@IanLee1521 So I'm okay with adding this check. That said, PEP-0008 is recommending this because it's a recommendation in PEP-0257 (which has [a check for this](https://github.com/GreenSteam/pep257/blob/master/pep257.py#L1074)). Perhaps we should recommend that tool in our documentation?

The problem we'll run into is needing an error code class if we do accept this. `E8` is wide open. But we might want to reserve that as "miscellany" in which we would throw this. An error class for anything that doesn't quite fit into any of the other classes. Alternatively, we could reach into hexadecimal to expand our error class possibilities. `=/`

SylvainDe:
Question and comments on various points :
 - as far as the error number is concerned, I have no strong opinion so I'll let you guys decide what you consider the best. On the same topic, I've arbitrarly made this an error. I might as well make it a warning if required. (I won't update the pull-request until this is decided in order not to waste everyone's time).
 - as for pep257, it is already mentionned in https://github.com/PyCQA/pep8/blob/master/docs/intro.rst but indeed, maybe it deserve a stronger mention/recommendation.
 - something I thought about for the triple-quoted string nested in triple-quoted strings (like the one @IanLee1521 commented for line 890) : maybe simple-quotes should be allowed for these situations as it keeps things more beautiful. One would just need to add a `and '"""' not in text` condition. Pros : more beautiful / more flexible. Cons : "Special cases aren't special enough to break the rules.". What do you think ?

sigmavirus24:
> something I thought about for the triple-quoted string nested in triple-quoted strings (like the one @IanLee1521 commented for line 890) : maybe simple-quotes should be allowed for these situations as it keeps things more beautiful. One would just need to add a and '"""' not in text condition. Pros : more beautiful / more flexible. Cons : "Special cases aren't special enough to break the rules.". What do you think ?

So, that's how pep257 already behaves. That makes me lean towards "we're reinventing something that another tool already does well" which makes me weary. Anyone then using `flake8` + `flake8-docstrings` will get 2x as many errors (perhaps not in agreement if done poorly) for the same error once we release this feature.

SylvainDe:
@sigmavirus24 So do you reckon this change should be abandoned ? I'm still in for having this check in vanilla pep8 as it is what I usually use but I don't want to work on this if we know it will bring more troubles than benefits.

sigmavirus24:
@SylvainDe this needs a rebase. 

I still haven't decided if we need to double up on checks between pep257 and pep8 but I'm not against this check's inclusion.

sigmavirus24:
The flake8 extensions that add pep257 support will probably just need to silence that check from pep257 by default.

SylvainDe:
@sigmavirus24 Thanks for the update! I've rebased my PR (or at least I've tried).

In any case, the error number is still wrong. I think it's easier for you to choose it than for me.
Also, do you think, I should send a PR to pep257 to silence that check by default ? (Note for later : https://github.com/GreenSteam/pep257/ )

sigmavirus24:
>  Also, do you think, I should send a PR to pep257 to silence that check by default ?

No. That's not what I was talking about. I'm talking about [PyCQA/flake8-docstrings](/pycqa/flake8-docstrings).

SylvainDe:
@sigmavirus24 Sorry for the delay in my answer. Thanks for the explanation ! I'll try to update this ASAP. Before I do anything, would you have an idea regarding the error number I should use (I'd rather fix this before messing with flake8-docstrings ).

SylvainDe:
@sigmavirus24 Before I fix all merge conflicts, do you have any suggestion for an error number ?

SylvainDe:
It seems like this has missed the 1.7 milestone. Anyway, does anyone have any suggestion regarding error number before I fix merge issues and try to send a pull request for PyCQA/flake8-docstrings as suggested above ?

sigmavirus24:
So I'm increasingly of the opinion that this check is satisfied by pep257 (soon to be renamed and moved into the PyCQA). That tool powers flake8-docstrings. I'm mostly worried that having two checks for people using both will not make for happy users. Also, the fact that as a project we're out of error classes to use for this is a little irksome to me in general.

asottile:
agreed, I believe that pep257 is the proper place for this check

PR #429, opened 2015-08-22
Multi-line ignore parsing in config_file

doismellburning:
In `tox.ini` it is entirely possible to have multi-line lists with trailing commas, e.g.

```
envlist =
    py27,
    py34,
```

If however you do

```
[pep8]
ignore =
    E1,
    E2,
```

then `options.ignore` ends up as `('\nE1', \nE2', '')`

The first two items of that tuple are "mildly inconvenient"; the preceding `\n` just means the ignore is...ignored - a consequence of the multiline

The empty string as the last item, a consequence of the trailing comma, is much more insidious, as it causes all checks to be silently ignored thanks to https://github.com/PyCQA/pep8/blob/master/pep8.py#L1890-L1901, specifically

```
return (code.startswith(self.options.ignore) and
                not code.startswith(self.options.select))
```

Currently, this PR includes some py27 tests to demonstrate this behaviour (I haven't tested my tests in other environments yet; am letting Travis do that for me) - I'll try to follow up with an actual fix.

Ultimately, the "real problem" here is probably "parsing another tool's config file"; there'll always be fun quirks etc., but that's much harder to do anything about!

### DIFF
--- a/pep8.py
+++ b/pep8.py
@@ -2081,10 +2081,10 @@ def process_options(arglist=None, parse_argv=False, config_file=None,
         options = read_config(options, args, arglist, parser)
         options.reporter = parse_argv and options.quiet == 1 and FileReport
 
-    options.filename = options.filename and options.filename.split(',')
+    options.filename = _parse_multi_options(options.filename.split(','))
     options.exclude = normalize_paths(options.exclude)
-    options.select = options.select and options.select.split(',')
-    options.ignore = options.ignore and options.ignore.split(',')
+    options.select = _parse_multi_options(options.select.split(','))
+    options.ignore = _parse_multi_options(options.ignore.split(','))
 
     if options.diff:
         options.reporter = DiffReport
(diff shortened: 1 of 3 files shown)


doismellburning:
(See https://github.com/doismellburning/cookiecutter-django/pull/7 for more context - many thanks to @rfleschenberg for his debugging assistance)

sigmavirus24:
I think there was already a bug for this, but it wasn't on the top of anyone's todo list. With tests already written, this should be a simple one for people to fix (I would hope).

doismellburning:
Welp, think I fixed it...

sigmavirus24:
:+1: Looks good to me.

sigmavirus24:
:sparkles: :cake: :sparkles: Thanks @doismellburning

doismellburning:
Cheers!

IanLee1521:
Thanks @doismellburning !

Issue #399, opened 2015-04-07
Consistent use of "return None" in functions


gvanrossum:
Could you add this rule to pep8? I just updated the official PEP 8 to include this (I'm surprised it wasn't already there):

* Be consistent in return statements. Either all return statements in a
  function should return an expression, or none of them should. If any return
  statement returns an expression, any return statements where no value is
  returned should explicitly state this as return None, and an explicit
  return statement should be present at the end of the function (if
  reachable).
```
   Yes:

   def foo(x):
       if x >= 0:
           return math.sqrt(x)
       else:
           return None

   def bar(x):
       if x < 0:
           return None
       return math.sqrt(x)

   No:

   def foo(x):
       if x >= 0:
           return math.sqrt(x)

   def bar(x):
       if x < 0:
           return
       return math.sqrt(x)
```
https://mail.python.org/pipermail/python-dev/2015-April/139054.html

IanLee1521:
Thanks, I saw this discussion and this helps remind me to implement it. :)

asottile:
fwiw, `mypy` is probably better suited to implement this type of check (and it already does):

```python
import math

def foo(x: float) -> float:
   if x >= 0:
       return math.sqrt(x)

def bar(x: float) -> float:
   if x < 0:
       return
   return math.sqrt(x)
```

```console
$ mypy --check-untyped-defs t.py
t.py:3: error: Missing return statement
t.py:9: error: Return value expected
Found 2 errors in 1 file (checked 1 source file)
```

SylvainDe:
Quite a lot of work has been spent on this issue but not much actual progress what integrated.
Before trying to go further, it may be interesting to have answers for the following questions:

1. do we still want to be handled by the pycodestyle tool ?

2. if so: do we agree that it needs to be an AST-based check ? An alternative could be to perform the check is a slighty weaken form as in https://github.com/PyCQA/pycodestyle/pull/711 which seems pretty interesting.

3. If so, as far as I can tell: no other check built in pycodestyle is based on AST check which raise various questions:

- should we ignore it by default to keep things fast by default ?

- should we first make sure that AST-based check can be implemented in the code base without much trouble ? (ie: ensuring that all the test suite is valid Python code as much as possible ? also: how to handle tests corresponding to code that is parsed as valid Python code depending on the Python version)


These questions come from discussions over various PRs:

- https://github.com/PyCQA/pycodestyle/pull/431 ( "Add checks for unconsistent returns" )

[...]

PR #431, opened 2015-08-25
Add checks for unconsistent returns

SylvainDe:
Added in PEP 8 :
"""
Be consistent in return statements. Either all return statements in a
function should return an expression, or none of them should. If any
return statement returns an expression, any return statements where no
value is returned should explicitly state this as return None, and an
explicit return statement should be present at the end of the function
(if reachable).
"""

Checking for unconsistent returns corresponds to implementing a check
based on ast tree. Given an AST tree, one can easily collect return
statements and check whether they return an explicit value or not.
Also, from the AST tree, one can try to check whether the end of
the function is reachable.

Warning W740 is added for explicit inconsistent returns.

Warning W741 is added for implicit inconsistent returns : when a
functions explicitly returns a value at some point but does not
at the (reachable) end of the function. If the end of a function
is not reachable but warning is triggered, one might simply add a
"return None" or an "assert False" : as one said : "Explicit is
better than implicit.".

Regarding the code :
Implementation has been made as easy to understand as possible.
The new check itself has been implemented in a new class
UnconsistentReturns which uses yet another class FlowAnalysis
which serves as an holder for various helper methods.
Also, I took this chance to change a few details so that AST-tree
checks fit more easily. This changes a few APIs. I don't know if
anyone is relying on those.

Regarding the tests :
Adding the first AST-tree based check leads to most of the
(incorrect) test code to be parsed which leads to many SyntaxError
either related to a single version of Python or to all of them. A
A new symbol has been to be able to convey the fact that an error
is expecting only for such or such version of Python.
I've fixed all issues related to SyntaxError so that they are all
passing all right. I hope I haven't changed what is actually tested.


SylvainDe:
Btw this is for #399

SylvainDe:
It seems like a few things do no pass properly on all Python versions. Will work on it soon.

sigmavirus24:
`pep8` is intended to be as fast as possible. Using the `ast` module defeats that purpose. This is why `pep8-naming` is a separate plugin. To avoid including ast parsing in pep8.

That said, however this is implemented, it should start out in the `DEFAULT_IGNORE` list.

SylvainDe:
I've updated the commit accordingly.
There is still a failure on Travis corresponding to test failing on pypy because of SyntaxError. I'll have to find the right way to fix this asap.

testsuite/W74.py
```diff
@@ -0,0 +1,101 @@
...
+    elif x == 0:
+        return 0
+#: W741:1:1
+def goldbach_conjecture_ko():
+    for i in itertools.count(2):
+        if not can_be_expressed_as_prime_sum(i):
+            return i
+
+
+# W741:1:1
```
SylvainDe: Confusing comment which is the glitch of a temporary version. No warning is actually returned (missing colon in the convent). To be removed asap.

SylvainDe:
Any news regarding this pull-request ?

SylvainDe:
Any news regarding this PR ?

ArcTanSusan:
@SylvainDe: There are branch conflicts that need fixing. I'm here at PyCon Sprints in Portland, OR. I'll be happy to fix the conflicts and propose a new branch.

rawrgulmuffins:
I'm going to try to fix the branch conflicts. My plan is to make my own fork, cherry pick these commits to that fork, rebase PyCQA master, and fix any conflicts that come up. Then I'll come back to this thread and reference my PR with the fixes.

rawrgulmuffins:
So having tried to get this PR to not have merge requests for a couple of hours I think that either the PR needs to be rebased against PyCQA Master (that will be hard since the PR merges master) or I can take the big idea's behind @SylvainDe's code and make my own branch.

I think I'd prefer to first since that gives @SylvainDe the commit credit. They most likely are going to have to do a `git reset --hard` on their last two commits, fetch PyCQA:master, and then rebase. 

If I haven't heard anything from them in the next couple of days I'll make me own branch and re-create their changes from a up to date fork.

IanLee1521:
A mid way point could be to 'got commit --author=...' Where you could recreate the attribution manually.

rawrgulmuffins:
I like that option actually. I can do this tomorrow when I sprint.

SylvainDe:
Hi guys,

I'm happy for this change to be integrated, whoever performs the final steps. In any case, I won't be able to do so in the next couple of days so if you guys want to make the best out of PyCon sprints, go for it. If you haven't time to do so, please let me know and I'll try to handle this as soon as possible.

Additional notes:
 - my commits/PR may contain changes which are not very relevant and can probably be discarded/pushed in a different commit (I was trying to wrap my head around the integration of AST checks at the time).
 - probably an irrelevant detail but in https://github.com/PyCQA/pycodestyle/pull/527 , the sentence "The original author is initially SylvainDe in #431, which is a PR that has been inactive for over a year" is just not true. If you look at the history, you have both commits and messages less than a year ago and until quite recently (?), this PR status was ready to be merged.

rawrgulmuffins:
@SylvainDe and @IanLee1521 I was told today by a friend at the PyCon sprints that using the AST in PyCodeStyle wasn't allowed. This change imports the AST.

Is that policy statement true?

If it is how should we approach this PR?

SylvainDe:
@rawrgulmuffins As far as I am concerned, I wish I knew the answer. From my point of view, this policy was made explicit only after I created the pull request (and I  think the exact text comes from the history you can find from this very page). My understanding is that the ast module will be actually used only if at least one check needs it so that you don't pay for what you don't need. Because the check for consistent return is not part of the default checks, I guess it's OK to have it rely on ast module because normal user won't see the performance impact. I'll let the official maintainer give you more details because that would much interest me (also I was wondering if there were any documentation about implementing ast checks as well). I hope the PyCon sprints are going well and I hope I could be there and enjoy it with you guys (and girls obviously).

rawrgulmuffins:
@SylvainDe Thanks for the quick reply. I've been having a lot of fun at the sprints this year.

SylvainDe:
@IanLee1521 @sigmavirus24 Now that I have a bit more free time on my hands, do you reckon it is worth spending some more time on this PR to solve the merge issues ?

FichteFoll:
This PR uses the word "unconsistent" in almost all places. As far as I'm concerned, [this is not a word](http://www.oxforddictionaries.com/spellcheck/english/?q=unconsistent) and "inconsistent" should be used.

SylvainDe:
Oh indeed. I'll fix this asap. Thanks for spotting.

rogalski:
@SylvainDe I intend to implement this check in Pylint, let me know if it's OK with you to reuse some of the code wrote by you in this PR.

SylvainDe:
It's totally fine for me :-)

Le 18 janv. 2017 23:30, "Łukasz Rogalski" <notifications@github.com> a
écrit :

SylvainDe:
@rogalski Also, feel free to ask me if anything at all needs some explanation.

SylvainDe:
The original PR is too big to be handled properly. I'll try to split into smaller, easier to integrate parts...

Issue #435, opened 2015-09-04
Regression in catching E711


ferulee46:
With the following code:
```
p[1] == '' or p[1] == None or p[0] == 0
```
pep8 v1.6.0 (and above) does not complain about `p[1] == None`
pep8 v1.5.7 does complain with: `E711 comparison to None should be 'if cond is None:'`

sigmavirus24:
So this regressed in order to fix false positives found in #330 and #336. The problem appears to be indexing into an iterable like is done in the example. If you check

```py
x == None
None == x
```

This works fine but as soon as you do

```py
x[0] == None
x() == None
```

It fails but curiously enough if you do:

```py
None == x[0]
None == x()
```

Then it works.

I suspect this is because #352 actually combined two regular expressions that were individually responsible for checking one direction (e.g., `x == None` vs `None == x`). Something was probably lost in translation.

IanLee1521:
Fixed with #442

PR #442, opened 2015-10-07
Fix regression in catching E711 (issue #435), add tests

tonkoandrew:


sigmavirus24:
If you roll back the change to the regular expression and run the tests, you absolutely do get test failures on lines: 

- 13
- 16
- 39
- 42

Which are tests that prove that the issue in #435 is fixed by this change. Well done @tonkoandrew

IanLee1521:
Great! Thanks @tonkoandrew ! And thanks @sigmavirus24 for the review!

PR #443, opened 2015-10-08
Attach the original source to the compiled tree

beck:
Issue:

When flake8 is being piped a file, stdin is consumed by pep8. A checker is
unable to evaluate the original source.

`seek(0)` is not possible since it is illegal in a pipe.  Constructing the
source from the tree gives unreliable line numbers.

The goal here is to build an IDE plugin, working specially with Atom.

### DIFF
--- a/pep8.py
+++ b/pep8.py
@@ -1510,7 +1510,9 @@ class Checker(object):
     def check_ast(self):
         """Build the file's AST and run all AST checks."""
         try:
-            tree = compile(''.join(self.lines), '', 'exec', PyCF_ONLY_AST)
+            source = ''.join(self.lines)
+            tree = compile(source, '', 'exec', PyCF_ONLY_AST)
+            tree.source = source
         except (ValueError, SyntaxError, TypeError):
             return self.report_invalid_syntax()
         for name, cls, __ in self._ast_checks:


pep8.py
```diff
@@ -1510,7 +1510,9 @@ def check_logical(self):
     def check_ast(self):
         """Build the file's AST and run all AST checks."""
         try:
-            tree = compile(''.join(self.lines), '', 'exec', PyCF_ONLY_AST)
+            source = ''.join(self.lines)
+            tree = compile(source, '', 'exec', PyCF_ONLY_AST)
+            tree.source = source
```
IanLee1521: Just running these lines at the command line (in Python 2.7.10) tree comes back as type `_ast.Module` which looks like it doesn't have a `source` attribute on it... Is this solely something that you make use of downstream to get back to the source in some fashion by monkeypatching the `source` onto the `tree`?

I think in either case I'd feel much more comfortable if an update to the docstring was added to clarify a bit.

asottile:
in flake8 3.x, you can request the `lines` parameter in your checker to retrieve this -- flake8 no longer uses `pycodestyle` as the underlying file execution engine

PR #446, opened 2015-10-17
Add optional getfullargspec for > PY36 compat

kevinconway:
The inspect.getargspec() function has been deprecated in PY3X and
finally removed in PY36. This patch adds optional use of the
inspect.getfullargspec() function which replaces it.

### DIFF
--- a/pep8.py
+++ b/pep8.py
@@ -1318,6 +1318,8 @@ def _get_parameters(function):
     if sys.version_info >= (3, 3):
         return list(inspect.signature(function).parameters)
     else:
+        if hasattr(inspect, 'getfullargspec'):
+            return inspect.getfullargspec(function)[0]
         return inspect.getargspec(function)[0]
 
 


sigmavirus24:
This is most certainly not where you want to be using that function. Note that you're adding that to the else-branch of an if-statement that checks that the version of Python pep8 is running on is greater than 3.3 (which includes, 3.6). In that case, it does the right thing because `getargspec` is deprecated. Otherwise, for older versions where that is not available, it uses `getargspec` which is fine.

kevinconway:
Yeah, I apologize for making this PR in haste. I'm going to close it.

I was hitting issues running the linter under 3.6 and didn't realize this fix was already in place but simply unreleased. I also wasn't paying attention to the surrounding code. A bit :disappointed: in myself for not doing my due diligence on this one.

Thanks for doing the code review and catching my mistake!

sigmavirus24:
No worries! I appreciate that you took the effort to fix it when you noticed it.

PR #455, opened 2015-10-28
optparse to argparse

higebu:
The optparse module is deprecated since Python 2.7, and I want use nargs in flake8 plugin.

### DIFF
--- a/pep8.py
+++ b/pep8.py
@@ -54,7 +54,8 @@ import time
 import inspect
 import keyword
 import tokenize
-from optparse import OptionParser
+import argparse
+from argparse import ArgumentParser
 from fnmatch import fnmatch
 try:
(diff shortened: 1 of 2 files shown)


sigmavirus24:
Please don't submit pull requests without first discussing the issue or problem in an issue or on [the mailing list](https://mail.python.org/mailman/listinfo/code-quality).

This library aims to support Python 2.6 and work as a standalone script with no dependencies. Python 2.6 usage still outnumbers Python 3 usage as a whole and it doesn't look like we will be dropping support for it in pep8 or flake8 anytime soon.

higebu:
I'm sorry but I missed the concept, "Small: Just one Python file, requires only stdlib. You can use just the pep8.py file for this purpose." in README.md.

PR #456, opened 2015-10-29
Correct _get_parameters.

jaccarmac:
The two versions of parameter parsing were not, in fact, equivalent. The
return types of `inspect.signature` and `inspect.getargspec` are, in
fact, subtly different.

See:
    * https://docs.python.org/3/library/inspect.html#inspect.signature
    * https://github.com/praw-dev/praw/issues/541#issuecomment-152280322

### DIFF
--- a/pep8.py
+++ b/pep8.py
@@ -1316,7 +1316,10 @@ _checks = {'physical_line': {}, 'logical_line': {}, 'tree': {}}
 
 def _get_parameters(function):
     if sys.version_info >= (3, 3):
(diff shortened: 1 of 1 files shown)


sigmavirus24:
This looks correct to me @IanLee1521. Thoughts?

jayvdb:
Looks good.

As an aside, there is talk of resurrecting inspect.getargspec() in Python 3.6: http://bugs.python.org/issue25486

IanLee1521:
Merged. Thanks!

PR #458, opened 2015-11-04
Do not report E302 occurring on first logical line

### LINKED ISSUE #412: Multiline comment before class at the beginning of file triggers E302
Possibly a duplicate of #174.

I may be misunderstanding something, but I understood this to be fixed a while ago. I'm using pep8 1.5.7. If I begin a file with

``` python
# comment comment comment comment comment comment comment comment comment comment
class Foo(object):
    pass
```

It says there is no syntax error. But if I wrap it to:

``` python
# comment comment comment comment comment comment comment comment
# comment comment
class Foo(object):
    pass
```

I get "E302 expected 2 blank lines, found 0" for the "class" line.

Was this fixed? Is there a config opt that turns it off? I couldn't find it in the docs.

jayvdb:
67b8be8 introduced an exception for E302 if the
first top level function or class was preceded
with only one comment line.

PEP 263 explicitly allows for two lines of comments
at the top of a module, in order to support a
hash-bang and an encoding declaration.

When the first top level function or class appears on the
first logical line, i.e. preceded only by comments, E302
should not be reported as the pep8 utility typically
ignores comments appearing anywhere in the code being checked.

Fixes #412

### DIFF
--- a/pep8.py
+++ b/pep8.py
@@ -254,7 +254,7 @@ def blank_lines(logical_line, blank_lines, indent_level, line_number,
     E303: def a():\n\n\n\n    pass
     E304: @decorator\n\ndef a():\n    pass
     """
-    if line_number < 3 and not previous_logical:
+    if not previous_logical and blank_before <= 2:
         return  # Don't expect blank lines before the first line
     if previous_logical.startswith('@'):
         if blank_lines:
--- a/testsuite/E30.py
+++ b/testsuite/E30.py
@@ -16,11 +16,6 @@ class X:
 #:
 
 
-#: E302:3:1
-#!python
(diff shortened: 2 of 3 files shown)


jayvdb:
IMO E303 (or a new code?) should also fire when there are two lines preceding normal code (such as `if`) occurring at the top level.
Happy to include that, or other similar improvements, on request.

sigmavirus24:
>  Happy to include that, or other similar improvements, on request.

PRs should be tiny and well focused. I think that idea would be best as a separate PR.

asottile:
oops didn't notice this, #919 includes a change which is ~essentially the same (slightly different in that it uses the configured top-level lines)

thanks for the patch!

PR #459, opened 2015-11-05
fix BytesWarning on Python 3

haypo:
On Python 3, when python3 with run with -bb, comparison between bytes
and str raises a BytesWarning exception. Test the Python version
instead of testing '' == b''.


haypo:
You can test simply with "python3 -bb pep8.py". Currently it fails with an exception.

sigmavirus24:
This looks great to me. Thanks @haypo

PR #469, opened 2016-01-11
v1.6.3 release

IanLee1521:
Review for release

@sigmavirus24 -- this should in theory solve the issues with flake8 compatibility, though if you hear anything there we should definitely fix it.

I'd like to see this release and any remaining issues with the flake8 integration worked out prior to doing the name change proposed in #466.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -2,14 +2,24 @@ Changelog
 =========
 
 
-1.6.x (unreleased)
+1.6.3 (unreleased)
 ------------------
+
+Announcements:
+
+* Repository moved to PyCQA Organization on GitHub:
+  https://github.com/pycqa/pep8
 
 Changes:
 
 * Reverted the fix in #368, "options passed on command line are only ones
   accepted" feature. This has many unintended consequences in pep8 and flake8
   and needs to be reworked when I have more time.
+* Added support for Python 3.5. (Issue #420)
+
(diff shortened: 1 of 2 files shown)


sigmavirus24:
I believe there have been far more changes included than would warrant a 1.6.3 release. If my memory serves me correctly this should actually be 1.7.0 and the changelog needs to include some more items. (@doismellburning, for example fixed some bugs around values parsed from config files.)

IanLee1521:
Not a bad idea (going to 1.7) I'll make that change tomorrow.

As an aside, I did skim the commit history since 1.6.2, must have missed a few. I'll add those tomorrow as well. 

Thanks for catching that @sigmavirus24

sigmavirus24:
So https://github.com/PyCQA/pep8/compare/1.6.2...HEAD shows more than what's in the changelog. Looks like:

- Python 3.5
- Fixing the config issues for Flake8
- Fixing parsing of comma-separated lists with new-lines
- Fix a BytesWarning on Python 3
- Fix how we get function signatures

IanLee1521:
Ah, I see what happened, I did the changelog mods on the release/v1.6.3 branch (https://github.com/PyCQA/pep8/compare/1.6.2...release/v1.6.3).

Going to close this and re-open on a new, release/v1.7 branch

Issue #484, opened 2016-02-24
Unary `-` reporting as W503


IanLee1521:
Copying from https://github.com/PyCQA/pycodestyle/issues/384#issuecomment-176760177 per @rkern.

> Similarly, unary `-` is falsely being reported as a W503 violation.
> 
> ```
> [~/scratch]$ cat false_pep8.py
> y = 75.0
> z = 100.0
> x = (y /
>      -z)
> [~/scratch]$ pep8 false_pep8.py
> false_pep8.py:4:6: W503 line break before binary operator
> [~/scratch]$ pep8 --version
> 1.7.0
> ```

sigmavirus24:
@rkern would you have time to try out #485?

PR #485, opened 2016-02-24
Identify binary operators used as unary operators

sigmavirus24:
Previously we handled the case where binary operators were being used
as unary operators except in the case where they followed another
binary operator, e.g.,

    foo = (1 +
           -10 *
           2 /
           -5)

This change updates the check for W503 to track the previous
non-whitespace token type and token text to check if it is in fact also
a binary operator (because you should never have two binary operators in
a row).

This does not handle invalid syntax, e.g.,

    foo = (bar /
           /baz)

But the false-positive generated for other cases was more harmful than
not catching what will instead be caught by the interpreter.

Closes gh-484

### DIFF
--- a/pep8.py
+++ b/pep8.py
@@ -1018,6 +1018,8 @@ def break_around_binary_operator(logical_line, tokens):
     Okay: foo(x,\n    -y)
     Okay: foo(x,  # comment\n    -y)
     Okay: var = (1 &\n       ~2)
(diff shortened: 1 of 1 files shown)


rkern:
Works for me! Thanks!

IanLee1521:
Thanks @sigmavirus24 + @rkern !

PR #486, opened 2016-02-25
Update Readme to reflect pycodestyle name change

jeffwidman:
Let me know if I missed anything... 

I assume the plan is to rename `pep8.py` to `pycodestyle.py`, but if not I can revert that part of the change.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,11 +1,16 @@
-pep8 - Python style guide checker
-=================================
+pycodestyle (formerly called pep8) - Python style guide checker
+===============================================================
 
-pep8 is a tool to check your Python code against some of the style
+pycodestyle is a tool to check your Python code against some of the style
 conventions in `PEP 8`_.
 
 .. _PEP 8: http://www.python.org/dev/peps/pep-0008/
 
+.. note::
+
+    This package used to be called ``pep8`` but was renamed to ``pycodestyle``
+    to reduce confusion. Further discussion `here
+    <https://github.com/PyCQA/pydocstyle/issues/466>`_.
 
 Features
 --------
@@ -15,18 +20,18 @@ Features
 * Parseable output: Jump to error location in your editor.
 
 * Small: Just one Python file, requires only stdlib. You can use just
-  the pep8.py file for this purpose.
+  the pycodestyle.py file for this purpose.
 
 * Comes with a comprehensive test suite.
 
 Installation
 ------------
 
-You can install, upgrade, uninstall pep8.py with these commands::
+You can install, upgrade, uninstall pycodestyle.py with these commands::
 
-  $ pip install pep8
-  $ pip install --upgrade pep8
-  $ pip uninstall pep8
+  $ pip install pycodestyle
+  $ pip install --upgrade pycodestyle
+  $ pip uninstall pycodestyle
 
 There's also a package for Debian/Ubuntu, but it's not always the
 latest version.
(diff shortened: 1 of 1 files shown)


sigmavirus24:
@IanLee1521 wrote https://github.com/PyCQA/pycodestyle/commit/2344a34a628b2f2e9df1c2539d9316005a35053d so I think he would have done this work at the same time if he meant to make any releases with pycodestyle yet. I suspect this is why `pep8.py` is not `pycodestyle.py` yet.

That said, I'm going to leave review in-line because you haven't updated the mark-up appropriately if @IanLee1521 does decide to accept this.

README.rst
```diff
@@ -1,11 +1,12 @@
-pep8 - Python style guide checker
+pycodestyle (formerly called pep8) - Python style guide checker
 =================================
```
sigmavirus24: This needs to be the same length as the title.

README.rst
```diff
@@ -1,11 +1,12 @@
...
 
-pep8 is a tool to check your Python code against some of the style
+pycodestyle is a tool to check your Python code against some of the style
 conventions in `PEP 8`_.
 
 .. _PEP 8: http://www.python.org/dev/peps/pep-0008/
 
+.. note:: This package used to be called :module:`pep8` but was renamed to :module:`pycodestyle` to reduce confusion. Further discussion here: `<https://github.com/PyCQA/pycodestyle/issues/466>`_
```
sigmavirus24: This should be wrapped at ~80 characters and when wrapping you need to do something like

```rst
.. note::

    This package used to be called ``pep8`` but was renamed to ``pycodestyle`` to
    reduce confusion. Further discussion `here 
    <https://github.com/PyCQA/pydocstyle/issues/466>`_.
```

For several reasons:

1. This is the README which is displayed on GitHub and PyPI. Neither will respect the `:module:` since neither uses sphinx
2. Attaching the link to `here` looks nicer.

IanLee1521:
That is the eventual plan, the discussion in #466 though argue for a longer deprecation period of ~ 1 year. Therefore some things like the pep8 vs pycodestyle packeages will transition more slowly to support users not switching over right away.

Some other issues that cropped up on the repo rename not called out in the #466 discussion, are (for instance) Read the Docs and Travis CI. Those will likely be handled in the next few weeks though.

sigmavirus24:
I think Travis CI is still working. If you add me to Read the Docs I can fix that for you. We can also set up `pycodestyle.pycqa.org` to be the documentation for this project so we don't have to worry about `pep8.readthedocs.org` versus `pycodestyle.readthedocs.org`.

jeffwidman:
Updated the PR based on helpful feedback from @sigmavirus24

README.rst
jeffwidman: fixed in latest commit

README.rst
jeffwidman: agreed on all points. I got bit by GitHub's web editor UI and didn't realize it wasn't wrapped. 

fixed in latest commit

README.rst
```diff
@@ -1,11 +1,16 @@
...
-pep8 is a tool to check your Python code against some of the style
+pycodestyle is a tool to check your Python code against some of the style
 conventions in `PEP 8`_.
 
 .. _PEP 8: http://www.python.org/dev/peps/pep-0008/
 
+.. note::
+
+    This package used to be called ``pep8`` but was renamed to ``pycodestyle``
+    toreduce confusion. Further discussion `here
```
sigmavirus24: to reduce (need a space) ;)

jeffwidman: Oops! Fixed

PR #487, opened 2016-02-28
Patched pep8 to comply with pydocstyle; Closes #228.

### LINKED ISSUE #228: pep8.py does not conform to pep-257
Running pep257.py on pep8.py results in several warnings:

- First line should end with a period.
- Blank line missing after one-line summary.
- Multiline docstring should end with 1 blank line.
- First line should be in imperative mood ('Do', not 'Does').
- One-liner docstrings should fit on one line with quotes.
- Class docstring should have 1 blank line around them.

IanLee1521:
Purposely ignoring D102 and D203 error codes.
- D102 appears to be a bug with D102
- D203 should be ignored by default by pydocstyle)

### DIFF
--- a/pep8.py
+++ b/pep8.py
@@ -837,7 +837,7 @@ def whitespace_before_comment(logical_line, tokens):
 
 
 def imports_on_separate_lines(logical_line):
-    r"""Imports should usually be on separate lines.
+    r"""Place imports on separate lines.
 
     Okay: import os\nimport sys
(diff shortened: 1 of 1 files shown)


sigmavirus24:
So far so good. I'm a little concerned though because the check docstrings (I think) are used by `--show-pep8` and typically lifted directly from the document. Perhaps we should find a better way to do that?

Issue #489, opened 2016-03-08
complain about missing space before opening parentheses of import statement


wbolster:
```python
from foo import(bar, baz)
```

should be

```python
from foo import (bar, baz)
```

but `pep8` does not currently complain about this. `import` is not a function!

sigmavirus24:
I can confirm this is a problem. I might be able to take a crack at fixing this later.

rogalski:
Shouldn't it be closed?

IanLee1521:
Yup. Thanks!

PR #491, opened 2016-03-23
complain about missing space before opening parentheses of import statement

rogalski:
Solves issue #489.

This checker may be better suitable for E2xx category (whitespace issues). Maybe even it should be connected to E221. I'll be happy to change if code review would settle for this alternative.


sigmavirus24:
I don't think this should be E403. This seems to fit better in with E229. Do you agree @IanLee1521?

IanLee1521:
I like E2xx better than E4xx, seems like it should be E275. I also wonder if the wording should be better as "missing whitespace after keyword" to match that family.

rogalski:
Hello, I've fixed commit based on your comments. Is any other action required from me right now?

sigmavirus24:
I don't think so. I'm :+1: on merging this.

IanLee1521:
Looks great! Thanks @rogalski !

PR #493, opened 2016-03-31
Add file name switch for proper handling of STDIN

leorochael:
Add `--stdin-name` switch for specifying a file name for the `-` file (STDIN).

Use the provided name for deciding if the file will be skipped.

Avoid calling `os.path.isdir()` and filename matching on the `-` filename.

Extensions, like flake8-putty, can use the provided name in name patterns,
which wasn't possible before.

See for example [this issue](https://github.com/AtomLinter/linter-flake8/issues/174).

The first commit of this pull request is optional.

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -1996,7 +1996,18 @@ class StyleGuide(object):
 
         Check if 'options.exclude' contains a pattern that matches filename.
         """
-        if not self.options.exclude:
+        result = self._excluded(filename, parent=parent)
+        if result and self.options.verbose:
+            print('ignoring excluded file: %s' % filename)
+        return result
+
+    def _excluded(self, filename, parent=None):
+        """Actual workhorse of the `.excluded()` method above.
+
+        See its docstring.
+        """
+        patterns = self.options.exclude
+        if not patterns:
             return False
         basename = os.path.basename(filename)
         if filename_match(basename, self.options.exclude):
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -1952,7 +1952,15 @@ class StyleGuide(object):
         report.start()
         try:
             for path in paths:
-                if os.path.isdir(path):
+                if path == "-":
+                    # STDIN
+                    path = self.options.stdin_name
+                    # if no ersatz filename for STDIN, assume not excluded
+                    excluded = path and self.excluded(path)
+                    if not excluded:
+                        runner(path or "stdin",
+                               lines=stdin_get_value().splitlines(True))
+                elif os.path.isdir(path):
                     self.input_dir(path)
                 elif not self.excluded(path):
                     runner(path)
@@ -2010,12 +2018,12 @@ class StyleGuide(object):
         if not patterns:
             return False
         basename = os.path.basename(filename)
-        if filename_match(basename, self.options.exclude):
+        if filename_match(basename, patterns):
             return True
         if parent:
             filename = os.path.join(parent, filename)
         filename = os.path.abspath(filename)
-        return filename_match(filename, self.options.exclude)
+        return filename_match(filename, patterns)
 
     def ignore_code(self, code):
         """Check if the error code should be ignored.
@@ -2095,6 +2103,9 @@ def get_parser(prog='pycodestyle', version=__version__):
     parser.add_option('--diff', action='store_true',
                       help="report changes only within line number ranges in "
                            "the unified diff received on STDIN")
+    parser.add_option('--stdin-name', metavar='FILE', default=None,
+                      help='File name to assume for when the file to check is'
+                           ' "-" (e.g. STDIN)')
     group = parser.add_option_group("Testing Options")
     if os.path.exists(TESTSUITE_PATH):
         group.add_option('--testsuite', metavar='dir',


pep8.py
```diff
@@ -1933,7 +1948,7 @@ def get_parser(prog='pep8', version=__version__):
     parser.config_options = [
         'exclude', 'filename', 'select', 'ignore', 'max-line-length',
         'hang-closing', 'count', 'format', 'quiet', 'show-pep8',
-        'show-source', 'statistics', 'verbose']
+        'show-source', 'statistics', 'verbose', 'stdin-name']
```
sigmavirus24: This doesn't make sense as something to put into a flake8 config file actually

leorochael: @sigmavirus24, True.

I was under the mistaken assumption that it was necessary for cmdline switches as well. Will fix.

leorochael:
There. Anything else I can do to speed merging this PR?

pep8.py
```diff
@@ -1890,7 +1898,14 @@ def excluded(self, filename, parent=None):
...
         Check if 'options.exclude' contains a pattern that matches filename.
         """
-        if not self.options.exclude:
+        result = self._excluded(filename, parent=parent)
+        if result and self.options.verbose:
+            print('ignoring excluded file: %s' % filename)
+        return result
+
+    def _excluded(self, filename, parent=None):
+        patterns = self.options.exclude
```
IanLee1521: Should probably update the other references in the function to `self.options.exclude` to use this `patterns` variable.

pep8.py
```diff
@@ -1890,7 +1898,14 @@ def excluded(self, filename, parent=None):
 
         Check if 'options.exclude' contains a pattern that matches filename.
         """
-        if not self.options.exclude:
+        result = self._excluded(filename, parent=parent)
+        if result and self.options.verbose:
+            print('ignoring excluded file: %s' % filename)
+        return result
+
+    def _excluded(self, filename, parent=None):
```
IanLee1521: Would you mind adding a docstring to this?

pep8.py
```diff
@@ -1846,7 +1846,15 @@ def check_files(self, paths=None):
...
             for path in paths:
-                if os.path.isdir(path):
+                if path == "-":
+                    # STDIN
+                    path = self.options.stdin_name
+                    # if no ersatz filename for STDIN, assume not excluded
+                    excluded = path and self.excluded(path)
+                    if not excluded:
+                        runner(path or "stdin",
+                               lines=stdin_get_value().splitlines(True))
```
IanLee1521: That is the rational for keeping the line ends?

IanLee1521:
A few minor comments, also needs to be rebased. I don't have any objections to merging this after the updates. Any other issues @sigmavirus24 ?

pep8.py
sigmavirus24: I'm curious about this too. Why are you keeping the line endings. Plus, why is this being special cased here?

pep8.py
```diff
@@ -1846,7 +1846,15 @@ def check_files(self, paths=None):
         report.start()
         try:
             for path in paths:
-                if os.path.isdir(path):
+                if path == "-":
+                    # STDIN
+                    path = self.options.stdin_name
```
sigmavirus24: What if someone didn't specify a name? In that case path is `None`. I'm guessing that will break things.

leorochael: If someone doesn't specify a name, then it defaults to the current behaviour. The name gets assigned to `'STDIN'`, or something similar (don't have it in front of me right now, but I manually tested that case).

sigmavirus24: Cool. @leorochael by the way, Flake8 3.0 beta releases support this for you. And it will be supported when 3.0 final is released as well. :)

pep8.py
leorochael: I just followed the lead of the `Checker` class, which contains this code in its `__init__()`:

```
        elif filename == '-':
            self.filename = 'stdin'
            self.lines = stdin_get_value().splitlines(True)
```

The reason it's being special cased here is that it's the first place in the code where we're examining the `path` variable, and it felt wrong to me to do the equivalent of `os.path.isdir('-')` as the first check.

the `isdir()` check usually fails and the `-` path would also fail the `excluded()` check and then be passed into the `Checker` and trigger the code above.

This assumes there isn't an actual `-` directory at the current working directory, so this PR actually fixes a hidden bug, I think (haven't tested this case).

On the other hand, if we pass in the lines explicitly into the checker (using the same code that `Checker` uses to split `stdin`), AND pass `None` as the filename, we hit this other branch of the `Checker` code:

```
        if filename is None:
            self.filename = 'stdin'
            self.lines = lines or []
```

Which means that `Checker` won't try to read the lines by itself, AND will assume the name is `stdin`.

Perhaps we should refactor this one line snippet somewhere, explain it and make it official, but I also don't know why `Checker` keeps the line endings :-)

Also, perhaps `Checker` should not be in the business of reading files and splitting lines, but this is not a refactor I have time to do right now :-)

pep8.py
leorochael: The explanation of the other question above shows exactly what happens when someone doesn't specify a name.

Yes, `path` will be `None`, but `Checker` is expecting it and will assume the file is name `stdin`.

pep8.py
leorochael: Since this method is the workhorse of `.exclude()` itself, I didn't feel the need at the time since it would just be repeating that method's docstring.

What do you suggest as docstring? Should we change the docstring of `.exclude()` to explain it's just a logging wrapper around `._exclude()` and move that docstring here?

leorochael:
@IanLee1521, @sigmavirus24, rebased and added docstring to `._exclude()`.

If you want a better docstring, just let me know...

asottile:
flake8 3.x supplies `--stdin-display-name` for this purpose (and no longer uses pycodestyle as the underlying file-running mechanism)

PR #502, opened 2016-04-30
Add W504 for line breaks before binary operators

sigmavirus24:
This flips the W503 rule to enforce line breaks before binary operators.

Related #498 

---

TODO:
- [x] Add W504 to list of codes
- [x] Add W504 to default ignore
- [x] Fix tests

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -6,6 +6,8 @@ UNRELEASED
 
 New checks:
 
+* Add W504 warning for checking that a break doesn't happen after a binar
+  operator. This check is ignored by default
 * Add W605 warning for invalid escape sequences in string literals
 
 2.3.1 (2017-01-31)
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -6,7 +6,7 @@ UNRELEASED
 
 New checks:
 
-* Add W504 warning for checking that a break doesn't happen after a binar
+* Add W504 warning for checking that a break doesn't happen after a binary
   operator. This check is ignored by default
 * Add W605 warning for invalid escape sequences in string literals
 
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -1184,16 +1184,15 @@ def break_before_binary_operator(logical_line, tokens):
 
     W503: (width == 0\n + height == 0)
     W503: (width == 0\n and height == 0)
+    W503: var = (1\n       & ~2)
+    W503: var = (1\n       / -2)
+    W503: var = (1\n       + -1\n       + -2)
 
-    Okay: (width == 0 +\n height == 0)
     Okay: foo(\n    -x)
     Okay: foo(x\n    [])
     Okay: x = '''\n''' + ''
     Okay: foo(x,\n    -y)
     Okay: foo(x,  # comment\n    -y)
-    Okay: var = (1 &\n       ~2)
-    Okay: var = (1 /\n       -2)
-    Okay: var = (1 +\n       -1 +\n       -2)
     """
     for context in _break_around_binary_operators(tokens):
         (token_type, text, previous_token_type, previous_text,
@@ -1215,25 +1214,26 @@ def break_after_binary_operator(logical_line, tokens):
 
     W504: (width == 0 +\n height == 0)
     W504: (width == 0 and\n height == 0)
+    W504: var = (1 &\n       ~2)
 
-    Okay: (width == 0\n + height == 0)
     Okay: foo(\n    -x)
     Okay: foo(x\n    [])
     Okay: x = '''\n''' + ''
     Okay: x = '' + '''\n'''
     Okay: foo(x,\n    -y)
     Okay: foo(x,  # comment\n    -y)
-    Okay: var = (1\n       & ~2)
-    Okay: var = (1\n       / -2)
-    Okay: var = (1\n       + -1\n       + -2)
+
+    The following should be W504 but unary_context is tricky with these
+    Okay: var = (1 /\n       -2)
+    Okay: var = (1 +\n       -1 +\n       -2)
     """
     for context in _break_around_binary_operators(tokens):
         (token_type, text, previous_token_type, previous_text,
          line_break, unary_context, start) = context
-        if (_is_binary_operator(previous_token_type, previous_text)
-                and line_break
-                and not unary_context
-                and not _is_binary_operator(token_type, text)):
+        if (_is_binary_operator(previous_token_type, previous_text) and
+                line_break and
+                not unary_context and
+                not _is_binary_operator(token_type, text)):
             error_pos = (start[0] - 1, start[1])
             yield error_pos, "W504 line break after binary operator"
 
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -1209,8 +1209,8 @@ def break_after_binary_operator(logical_line, tokens):
     r"""
     Avoid breaks after binary operators.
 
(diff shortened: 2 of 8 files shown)


sigmavirus24:
@IanLee1521 I know the tests are failing, and I'm going to have to figure out how to fix them, but I'd love some code review.

pep8.py
```diff
@@ -1018,7 +1018,15 @@ def explicit_line_join(logical_line, tokens):
                 parens -= 1
 
 
-def break_around_binary_operator(logical_line, tokens):
+def _is_binary_operator(token_type, text):
+    # The % character is strictly speaking a binary operator, but the
```
IanLee1521: Minor, but I'd prefer to move this to a `"""..."""` style doc string rather than `#` since we're touching it anyways.

pep8.py
```diff
@@ -1018,7 +1018,15 @@ def explicit_line_join(logical_line, tokens):
...
+    # common usage seems to be to put it next to the format parameters,
+    # after a line break.
+    return ((token_type == tokenize.OP or text in ['and', 'or']) and
```
IanLee1521: Would there be any objections to changing this line (and in the future other similar lines) to something like:

```python
token_op = token_type == tokenize.OP
and_or = text in ['and', 'or']
not_symbol = text not in "()[]{},:.;@=%~"

return ((token_op or and_or) and not_symbol)
```

Not only does this make the line wrapping issue go away, but I think it makes it a bit clearer what is going on. Note: These are just sample variables names, maybe there are even better names for them.

pep8.py
```diff
@@ -1056,17 +1057,61 @@ def is_binary_operator(token_type, text):
...
+                    not _is_binary_operator(previous_token_type,
+                                            previous_text)):
                 yield start, "W503 line break before binary operator"
             unary_context = text in '([{,;'
             line_break = False
             previous_token_type = token_type
             previous_text = text
 
 
+def break_after_binary_operator(logical_line, tokens):
```
IanLee1521: My one major twitch with this is that now we have two functions that are nearly identical... Is there something we could do to factor out the commonality between them?

IanLee1521:
@sigmavirus24 -- Made a few in line comments, nothing major, overall looks good to me 👍 if we fix the tests I have no issues merging.

pep8.py
sigmavirus24: Yeah that's on my road map for this. It makes me yearn for Python 3's `yield from` because it would be much cleaner than creating one generator that both iterate over and then also yield results from. :/

IanLee1521: So, I haven't followed the `yield from` functionality all that closely, but if you don't mind, could you give an example of what this would look like with that feature? Getting a bit out of the scope of the PR, but might be something we come back to in the future.

sigmavirus24: So let me give you a practical example, let's say we have a function that takes a callable and yields data and it looks like `gen_func_0(filter=None)`. Let's say we have a common filter that we always want to apply and just return what's there, that would be a good use case for another function, e.g., `gen_func_1()` which wants to yield data yielded by `gen_func_0`. In Python 2, `gen_func_1` is implemented like so:

```py
def gen_func_1():
    for result in gen_func_0(common_filter):
        yield result
```

or 

```py
def gen_func_1():
    return (r for r in gen_func_0(common_filter))
```

The latter is shorter, but neither is really elegant. In Python 3, you could instead do:

```py
def gen_func_1():
    yield from gen_func_0(common_filter)
```

This means that in a magical future where that function becomes part of an asyncio application you can also then decorate it as something that is a co-routine (which doesn't help us in this very case, but makes it a good feature to start using).

IanLee1521:
@sigmavirus24 -- I've been working most of the day on fixing the tests and getting some other things merged. I captured the branch at: https://github.com/PyCQA/pycodestyle/tree/gh-502

I'm on the last failure (https://travis-ci.org/PyCQA/pycodestyle/builds/140304242) still chipping away if you have any thoughts...

hoylemd:
Are there any updates on this? My team would very much like to use it.  If development has stalled,  could we grab the branch and polish it off?

sigmavirus24:
@hoylemd, thanks for the interest! @IanLee1521 has updated this a bit on https://github.com/PyCQA/pycodestyle/tree/gh-502 but there are still failing tests. If you can fix those, you're welcome to take over this work and submit a new PR!

peterjc:
It looks like flake8 assumed this was going to be merged and already ignores it by default?
https://gitlab.com/pycqa/flake8/issues/397

FichteFoll:
The only thing holding this back are failing tests, right?

hoylemd:
Well, it probably needs review too, but tests should come first.  @FichteFoll Are you planning to take a run at the tests?

FichteFoll:
I was cosidering it, but I have enough tasks in my backlog already that it'd probably take a long time and I don't want to block others from going at it.

sigmavirus24:
@IanLee1521 could you lend a review? Also @hoylemd and @peterjc could either of you verify this independently? I know there's still a nasty edge-case where it won't catch the problem but right now this is good enough.

CHANGES.txt
```diff
@@ -6,6 +6,8 @@ UNRELEASED
 
 New checks:
 
+* Add W504 warning for checking that a break doesn't happen after a binar
```
IanLee1521: typo: `binar` -> `binary`

pycodestyle.py
```diff
@@ -1174,17 +1166,78 @@ def is_binary_operator(token_type, text):
...
+                                        previous_text)):
+            yield start, "W503 line break before binary operator"
+
+
+@register_check
+def break_after_binary_operator(logical_line, tokens):
+    r"""
+    Avoid breaks after binary operators.
+
+    The preferred place to break around a binary operator is after the
```
IanLee1521: Copy/paste error -- This docstring should be `... is before the operator, not after it.`

sigmavirus24:
>For some reason I can't edit the files... Do you have the "allow contributions from maintainers" disabled?

That's enabled now. Sorry about the copy-pasta fail and botching the CHANGES file. I can fix those up tonight

IanLee1521:
No worries. I just committed those fixes. Should impact the build, but will wait for it to pass again, then I'll merge.

IanLee1521:
Thanks for the work on this @sigmavirus24 !

hoylemd:
@sigmavirus24 Ah, sorry, I didn't see this until today.  Next time, I hope!

FichteFoll:
Thanks guys ❤️ . This is my preferred style and I love forcing contributors to adhere to it.

peterjc:
OK, the PR was updated and merged while I was away from the computer, so better late than never? Testing from the master branch:

```
$ git clone https://github.com/PyCQA/pycodestyle.git
$ cd pycodestyle
$ git checkout 534310014a1dad2a5aeca486fa2dccda93eb930b
$ python3 -m pip install .
```

After than I used ``python3 -m pycodestyle --select W504 ...`` on some large code bases and it gave sensible results. Thanks everyone.

peterjc:
New code ``W504`` needs adding to https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes - is this just a case of updating the following RST file?

https://github.com/PyCQA/pycodestyle/blob/master/docs/intro.rst#error-codes

sigmavirus24:
@peterjc yes, that would be the file to update.

peterjc:
Pull request #724 submitted to update the list of error codes.

Issue #507, opened 2016-05-23
False positive E251 about spaces around "=" for annotated parameters with default values in async functions 


east825:
According to PEP8 and #357 it's allowed to wrap with spaces equal sign before the default value of a parameter if the latter has annotation, e.g. `x: int = 42`. However if containing function is also marked as `async`, pep8 still falsely reports E251 "unexpected spaces around keyword / parameter equals" for its parameters.

For instance, for the next function pep8.py complains about whitespaces surrounding "=" sign. Meanwhile, if you remove `async` keyword, it now longer does.

```python
async def func(x: int = 42):
    pass
```

sigmavirus24:
So https://github.com/PyCQA/pycodestyle/pull/361 fixed it (without the async keyword) *before* the async keyword was added. We can probably just update that fix to allow this to work properly.

sigmavirus24:
I have something [here](https://github.com/PyCQA/pycodestyle/compare/master...bug/507) but GitHub isn't let me make a pull request bizarrely.

sigmavirus24:
#508 should fix this.

east825:
@sigmavirus24 Great! Thank you.

PR #508, opened 2016-05-23
Allow spaces around = in async definitions

sigmavirus24:
In pull request gh-361, we allowed spaces around = for default
arguments with annotations. Python 3.5 added the async keyword for
a function definition and the allowance we made in gh-361 was failing.

This allows a function definition to start with either 'def' or
'async def' now and accommodates both cases.

Closes gh-507


IanLee1521:
Thanks @sigmavirus24 !

PR #512, opened 2016-05-25
Fix spelling

jwilk:
These misspellings were found using [mwic](http://jwilk.net/software/mwic).


sigmavirus24:
Thanks @jwilk

IanLee1521:
Great! Thanks! I'll have to check out that tool...

PR #522, opened 2016-06-02
Updated pull request #314 to current master.

arthall:
Updated pull request #314 to current master, now passes tests in E73.py.


IanLee1521:
Great, thanks @arthall and @darioush ! These are the timing results for a 10k element dict on my laptop for the record:

```
In [25]: s = str({str(x):x for x in range(10000)})

In [26]: %timeit list(pycodestyle_before.compound_statements(s))
1 loops, best of 3: 762 ms per loop

In [27]: %timeit list(pycodestyle_after.compound_statements(s))
100 loops, best of 3: 9.81 ms per loop
```

Issue #341, opened 2014-10-31
Badly named single-letter variable names are not detected


ThiefMaster:
```shell
> cat test.py
def l(I):
    l = O = l = 0 + I
> pep8 test.py
>
```

> Never use the characters 'l' (lowercase letter el), 'O' (uppercase letter oh), or 'I' (uppercase letter eye) as single character variable names.

> In some fonts, these characters are indistinguishable from the numerals one and zero. When tempted to use 'l', use 'L' instead.

IanLee1521:
For the record, this is from: https://www.python.org/dev/peps/pep-0008/#names-to-avoid

PR #525, opened 2016-06-02
Detect single letter variable names enumerated in PEP-0008

mtmiller:
Fix for #341. Could be improved further to include named parameters, maybe other cases I can't think of where names are declared.


pycodestyle.py
```diff
@@ -1150,6 +1150,37 @@ def comparison_type(logical_line, noqa):
...
 
 
+def ambiguous_identifier(logical_line, tokens):
+    r"""Never use the characters 'l', 'O', or 'I' as variable names.
+
+    In some fonts, these characters are indistinguishable from the numerals
+    one and zero. When tempted to use 'l', use 'L' instead.
+
+    Okay: L = 0
+    E741: l = 0
```
IanLee1521: Thanks @mtmiller ! Could you update the doc tests with the rest of the cases that are not allowed?

Also, the 'as' test cases are missing.

mtmiller:
Added checks for names after `class`, `def`, `global`, and `nonlocal`, with new error IDs for function and class definitions versus local variables.

Now reports errors for almost everything in the original example in #341.

Still missing checks for for loop iterators, function parameters, and the comma-separated lists of variables that can appear in assignments, for loops, arg lists, and `global` and `nonlocal` statements.

sigmavirus24:
This will have to be included in 2.1.0 or some other 2.x.0 release.

IanLee1521:
Next version will be 2.1.0, going to go ahead and merge this. Thanks @mtmiller !

PR #528, opened 2016-06-02
For Issue 293

alecklandgraf:
Adds test for `E133` which is enabled by the optional flag `--hang-closing`. 

Per the docs: 

> The check E133 is mutually exclusive with check E123.

Specifically this throws an `E133` error for 

```py
a = [
    12
]
```

and is Okay for

```py
a = [
    12
    ]
```


sigmavirus24:
I'm not quite sure I like this approach but I'm also not sure there's a better approach. :(

testsuite/support.py
```diff
@@ -146,6 +146,19 @@ def init_tests(pep8style):
     report = pep8style.init_report(TestReport)
     runner = pep8style.input_file
 
+    def set_hang_closing_option(pep8style, codes):
+        """Set hang-closing option if present in the test codes.
+
+        (mutates pep8style)
+
+        Usage in test:
+        #: #E133:3:1 --with-hanging
```
sigmavirus24: This is wrong, the usage in test would be:

```
#: E133:3:1 --with-hanging
```

alecklandgraf: Reading it over, I got the command line arg wrong too, I'll update this, should be ``--hang-closing`` not ``--with-hanging``

alecklandgraf:
I'm open to other approaches, my other thought was to have another test file, e.g. `test_flags.py` which could test over a set of expected errors, but that would break the current structure of having all E12X in the `E12.py` file, and might be too much overhead to keep track of.

Let me know if you'd like me to rebase.

PR #529, opened 2016-06-02
Fix cross-platfrom issues with test suite

jnmclarty:
Several minor issues related to windows inhibit a smooth development life cycle.

The commits are labelled appropriately with the changes.  They are independent.

Notice the '*' logic is ignored isn't tested on windows.  That edge case (feature) may or may not work depending on minor release of python.  Testing for it, is more hassle.

This was tested locally against 2.7, 3.4 and 3.5.  I'm counting on travis for the rest.

I have a more strategic version of the normalize function, which I may finish later and send after this is merged.


jnmclarty:
Ian, the risk of this PR is much lower risk than all the solution paths we discussed before you left.  I erred on the side of fixing the test suite, without attempting cleverness, rather than implementing a strategic solution which would break be a backwards incompatibility nightmare.  Right now, the folders given by users, are allowed not to exist.  So, all the common checks, involving checking if a string is a folder, rely on existence rather than say, a regex based approach.

testsuite/test_shell.py
```diff
@@ -7,6 +7,26 @@
...
+def safe_line_split(line):
+    """ parses the path, from message details """
+
+    # On certain platforms, the drive notation conflicts
+    # with the message seperator, inducing problems during tuple unpacking.
+
+    split = line.split(':')
+    if 'win' in sys.platform:
+        if len(split) == 5:
+            drive, path, x, y, msg = split
```
sigmavirus24: Can you leave comments explaining what kind of situation would result in having 5 items? (Likewise for the elif checking for a length of 4) Bonus points for an example in the comment.

IanLee1521: Agreed, that will be helpful as we continue to build up the windows support and verify that use case.

testsuite/test_shell.py
```diff
@@ -7,6 +7,26 @@
 from testsuite.support import ROOT_DIR, PseudoFile
 
 
+def safe_line_split(line):
+    """ parses the path, from message details """
+
+    # On certain platforms, the drive notation conflicts
```
IanLee1521: Also, these comments can / should probably just move into the function docstring.

jnmclarty:
Hi Folks, it appears your comments slipped through my inbox. I'll get to the feedback on the weekend.

jnmclarty:
Merge?

jayvdb:
The changes to `normalize_paths` are a breaking change on Windows, and it looks like they will break https://github.com/jayvdb/flake8-putty , which converts `os.sep` to `/` because that is what pycodestyle handles correctly. (https://github.com/jayvdb/flake8-putty/commit/a0c7c604615)

To avoid this being a breaking change, `normalize_paths` needs to handle `os.sep` and [`os.altsep`](https://docs.python.org/3.0/library/os.html#os.altsep) on Windows.

Then your changes to `testsuite/test_util.py` are not needed.

sigmavirus24:
> The changes to normalize_paths are a breaking change on Windows

Can you explain why they're a breaking change? Does it break plugins or something else?

jayvdb:
No. Command line usage also breaks. The changes to the test suite show how it breaks.

Issue #518, opened 2016-06-01
--help still shows pep8 for Usage


sgornick:
When I view the help I still see "Usage: pep8 [options] input ...".

Wouldn't pep8 be something that is no longer referenced, and pycodestyle shown instead?

```
$ pycodestyle --help
Usage: pep8 [options] input ...

Options:
  --version            show program's version number and exit
  -h, --help           show this help message and exit
[...]

$ pycodestyle --version
2.0.0
```

IanLee1521:
Yep. Good catch! I'll probably leave that as a bug that can be fixed at the PyCon sprint on Thursday. Would be a great first contribution by someone!

SylvainDe:
Also, wiki still has references to pep8 : https://github.com/PyCQA/pycodestyle/wiki . Not sure if better to 
have one bug open to track all these issues or one bug per issue.

vodik:
Got it.

IanLee1521:
Thanks for that catch @SylvainDe -- Let's split that out into a separate issue, primarily because I'm not sure that wiki content should be kept around at all.. More when I open the new issue in a minute...

IanLee1521:
@SylvainDe -- See #531

IanLee1521:
Thanks for the report @sgornick, this is fixed with the merging of #530.

PR #519, opened 2016-06-01
change help name from pep8 to pycodestyle

xiaopeng163:
#518 
Signed-off-by: Peng Xiao <xiaoquwl@gmail.com>

PR #530, opened 2016-06-02
Change all references to the pep8 project to say pycodestyle

iamaspacecow:
This fixes issue #518, pep8 still referenced in the cli help command
As a side effect,  in setup.cfg now becomes
Also, changed the path for the config file from ~/.config/pep8 to ~/.config/pycodestyle
These feel like changes that should have come with the jump to version 2.0.0, as they are breaking,
but support for  as a name can still be added if it's desired enough

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -167,13 +167,14 @@ Quick help is available on the command line::
       --benchmark        measure processing speed
 
     Configuration:
-      The project options are read from the [pep8] section of the tox.ini
-      file or the setup.cfg file located in any parent folder of the path(s)
-      being processed.  Allowed options are: exclude, filename, select,
+      The project options are read from the [pycodestyle] section of the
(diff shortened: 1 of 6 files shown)


ArcTanSusan:
There's an incomplete sentence in commit message: `As a side effect, in setup.cfg now becomes`

ArcTanSusan:
is this PR intended to cover every mention of `pep8` in all docs, code, tests? https://gist.github.com/c1b47cf96a899786c3b0b92501d79297 is the output of "git grep pep8" on your branch.

sigmavirus24:
How does this differ from #519 and #535?

xiaopeng163:
your PR is better than mine, I will close #519 , thanks

IanLee1521:
Looks great, thanks @iamaspacecow ! Also thanks to @xiaopeng163 and @ArcTanSusan for your feedback and input.

Closes #535

PR #532, opened 2016-06-03
Update Python version list in docs

mtmiller:
Just a tiny doc update to list the correct Python versions that are currently tested in Travis.


sigmavirus24:
Might as well just say "Latest stable version of Python 3"

ArcTanSusan:
👍 to avoid having to update that version number on every new release

mtmiller:
Fair enough, would be simpler.

OTOH `tox.ini` has to be updated for every new Python release anyway :wink:

IanLee1521:
Thanks @mtmiller !

PR #534, opened 2016-06-03
Fix typos

ArcTanSusan:


testsuite/test_api.py
```diff
@@ -138,7 +138,7 @@ def test_styleguide(self):
         self.reset()
 
     def test_styleguide_options(self):
-        # Instanciate a simple checker
+        # Instanstiate a simple checker
```
JelleZijlstra: Instantiate

pycodestyle.py
```diff
@@ -120,7 +120,7 @@
 LAMBDA_REGEX = re.compile(r'\blambda\b')
 HUNK_REGEX = re.compile(r'^@@ -\d+(?:,\d+)? \+(\d+)(?:,(\d+))? @@.*$')
 
-# Work around Python < 2.6 behaviour, which does not generate NL after
+# Work around Python < 2.6 behavior, which does not generate NL after
```
sigmavirus24: This is actually a difference between American English and British English. I don't consider it a typo.

testsuite/test_api.py
sigmavirus24: Once you update this and revert the change to behaviour this will be good to go.

pycodestyle.py
IanLee1521: Agreed.

PR #539, opened 2016-06-07
Update maximum_line_length to use Checker.noqa

### LINKED ISSUE #538: maximum_line_length determines NOQA the hard way
The Checker has a `noqa` attribute that is updated per-physical line. This can be requested by a check in its signature. `maximum_line_length`, however, runs the regular expression a second time on the line instead of relying on the Checker attribute. Regular expressions in Python aren't exactly the fastest, so repeatedly doubling the work is bad for performance.

sigmavirus24:
This also updates the Checker to update it's state for physical lines
as well as logical lines. This allows maximum_line_length to rely on
Checker state for its noqa check.

Closes #538

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -199,7 +199,7 @@ def trailing_blank_lines(physical_line, lines, line_number, total_lines):
             return len(physical_line), "W292 no newline at end of file"
 
 
-def maximum_line_length(physical_line, max_line_length, multiline):
+def maximum_line_length(physical_line, max_line_length, multiline, noqa):
     r"""Limit all lines to a maximum of 79 characters.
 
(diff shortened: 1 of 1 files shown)


IanLee1521:
Thanks @sigmavirus24

jayvdb:
A relnote would be good for this as many flake8 plugins could benefit from it, as they often do thier own noqa support.

sigmavirus24:
@jayvdb I'm not sure which plugins do that, but this attribute has been on the Checker for as long as I can remember and has been used by hacking for *years*.

Either way, plugins for Flake8 will be changing drastically in 3.0 so plugins will not have to do their own noqa check.

PR #541, opened 2016-06-07
Switch Travis to use tox

sigmavirus24:
This reduces confusion about how travis runs tests versus a local
developer.

**Note** this also adds a Flake8 test env and runs it at the gate to catch other instances of https://github.com/PyCQA/pycodestyle/pull/540

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,23 +1,28 @@
 language: python
 sudo: false
-python:
-  - 2.6
-  - 2.7
-  - 3.2
-  - 3.3
-  - 3.4
-  - 3.5
-  - nightly
-  - pypy
-  - pypy3
 install:
-  - pip install -e .
-  - pip list
-script:
-  - python pycodestyle.py --testsuite testsuite
-  - python pycodestyle.py --statistics pycodestyle.py
-  - python pycodestyle.py --doctest
-  - python setup.py test
+  - pip install tox
+script: tox
+matrix:
+  include:
+    - python: 2.6
+      env: TOXENV=py26
+    - python: 2.7
+      env: TOXENV=py27
+    - python: 3.3
+      env: TOXENV=py33
+    - python: 3.4
+      env: TOXENV=py34
+    - python: 3.5
+      env: TOXENV=py35
+    - python: 3.6
+      env: TOXENV=py36
+    - python: nightly
+      env: TOXENV=py37
+    - python: pypy
+      env: TOXENV=pypy
+    - python: 3.5
+      env: TOXENV=flake8
 
 notifications:
   email:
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -14,3 +14,10 @@ commands =
 	{envpython} pycodestyle.py --statistics pycodestyle.py
 	{envpython} pycodestyle.py --doctest
 	{envpython} -m testsuite.test_all
+
+
+[testenv:flake8]
+deps =
+    flake8
+commands =
+    flake8 pycodestyle.py
```
IanLee1521: What's the rationale behind running flake8 on pycodestyle? Not specifically disagreeing, just seems to not be part of the "Switch Travis to use tox" issue.

.travis.yml
```diff
@@ -1,23 +1,29 @@
...
-  - pip install -e .
+  - pip install tox
   - pip list
-script:
-  - python pycodestyle.py --testsuite testsuite
-  - python pycodestyle.py --statistics pycodestyle.py
-  - python pycodestyle.py --doctest
-  - python setup.py test
+script: tox
+matrix:
```
IanLee1521: Why convert from the python to the matrix version of this?

.travis.yml
```diff
@@ -1,23 +1,29 @@
...
 install:
-  - pip install -e .
+  - pip install tox
   - pip list
-script:
-  - python pycodestyle.py --testsuite testsuite
-  - python pycodestyle.py --statistics pycodestyle.py
-  - python pycodestyle.py --doctest
-  - python setup.py test
+script: tox
```
IanLee1521: I like this part! I had come up with the same idea and forgotten to change it... Great minds! :)

.travis.yml
jayvdb: It can be avoided using `tox-travis`

sigmavirus24: @IanLee1521 it matches the usage of the majority of projects that have test environments that are more than just "test on these versions of python". It would allow the project to, for example, also run a tox test environment to ensure that the documentation passes a tool like doc8 and builds in sphinx with all warnings treated as errors, etc. (Effectively testing that your docs meet a certain style and actually build.)

Given how many more projects use the matrix approach rather than using the rather young tox-travis tool (which I've not seen any project of any size use), I think this is going to be the better way to manage our usage of tox within travis.

tox.ini
sigmavirus24: Flake8, as I'm sure you know, includes pyflakes. When I was writing some of the more recent pull requests, I found several pyflakes lint problems (which I also submitted separate pull requests for). If pycodestyle runs flake8 on all submitted changes, then those shouldn't pop up again.

It's something of an extra check to ensure the quality of pycodestyle doesn't dip.

PR #542, opened 2016-06-08
Avoid redefining path in the list comprehension

sigmavirus24:
Use a distinct name in our parsing of the unified diff instead.

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -198,7 +198,7 @@ def trailing_blank_lines(physical_line, lines, line_number, total_lines):
             return len(physical_line), "W292 no newline at end of file"
 
 
-def maximum_line_length(physical_line, max_line_length, multiline):
+def maximum_line_length(physical_line, max_line_length, multiline, noqa):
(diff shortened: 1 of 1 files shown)


jayvdb:
IMO it is the `path` in the three line dict comprehension at the end which should be changed, rather than breaking blame of the more complex code unified diff parsing loop.

Also the path isnt 'parsed'; it is 'extracted' with a standardised prefix cut off.

sigmavirus24:
I can sit here and bike-shed over variable names, but that's not useful for either of us @jayvdb. As for which instance of "path" to change, I don't really care.  I don't think using a longer variable name will make the comprehension any easier to read, so I'm merely using a more descriptive name elsewhere. The other sad part is that I understand that function quite well. So if someone wants to ping me about it because my name shows up in the blame, they're more than welcome to do it.

jayvdb:
Another approach is to add `del path` between the two chunks of code.

sigmavirus24:
> Another approach is to add del path between the two chunks of code.

I don't see how that's better than providing the local variable with a more descriptive name.

Issue #506, opened 2016-05-18
Some better handling of error in the configuration file


mcepl:
So, I have left some merge conflict markers in my ``setup.cfg`` and this how the pep8 (run through flake8) dealt with it:
```
ConfigParser.py:546:_read:ParsingError: File contains parsing errors: /home/matej/archiv/knihovna/repos/m2crypto/setup.cfg
	[line  2]: '<<<<<<< 642f88cb1b6027e184d9a662b255f7fea4d9eacc\n'
	[line  9]: '=======\n'
	[line 12]: '>>>>>>> Open trunk for 0.22.\n'

Traceback (most recent call last):
  File "/usr/bin/flake8", line 9, in <module>
    load_entry_point('flake8==2.0', 'console_scripts', 'flake8')()
  File "/usr/lib/python2.7/site-packages/flake8/main.py", line 21, in main
    flake8_style = get_style_guide(parse_argv=True, config_file=DEFAULT_CONFIG)
  File "/usr/lib/python2.7/site-packages/flake8/engine.py", line 76, in get_style_guide
    styleguide = StyleGuide(**kwargs)
  File "/home/matej/.local/lib/python2.7/site-packages/pep8.py", line 1788, in __init__
    arglist, parse_argv, config_file, parser)
  File "/home/matej/.local/lib/python2.7/site-packages/pep8.py", line 2074, in process_options
    options = read_config(options, args, arglist, parser)
  File "/home/matej/.local/lib/python2.7/site-packages/pep8.py", line 1999, in read_config
    if config.read(os.path.join(parent, fn) for fn in PROJECT_CONFIG):
  File "/usr/lib64/python2.7/ConfigParser.py", line 305, in read
    self._read(fp, filename)
  File "/usr/lib64/python2.7/ConfigParser.py", line 546, in _read
    raise e
ParsingError: File contains parsing errors: /home/matej/archiv/knihovna/repos/m2crypto/setup.cfg
	[line  2]: '<<<<<<< 642f88cb1b6027e184d9a662b255f7fea4d9eacc\n'
	[line  9]: '=======\n'
	[line 12]: '>>>>>>> Open trunk for 0.22.\n'

Local variables in innermost frame:
sectname: 'flake8'
cursect: OrderedDict([('__name__', 'flake8'), ('ignore', ['E402,N806,N803,N802,E501']), ('tag_build', ['.dev']), ('tag_svn_revision', ['1'])])
[...]

PR #547, opened 2016-06-16
Handle ParsingErrors from configparser

sigmavirus24:
In the event that the config file is mangled, we should should be able
to recover from that without issue.

Closes gh-506

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -60,10 +60,10 @@ from fnmatch import fnmatch
 from optparse import OptionParser
 
 try:
-    from configparser import RawConfigParser
+    import configparser
     from io import TextIOWrapper
 except ImportError:
-    from ConfigParser import RawConfigParser
+    import ConfigParser as configparser
 
 __version__ = '2.1.0.dev0'
 
@@ -2092,7 +2092,7 @@ def read_config(options, args, arglist, parser):
     configurations in the current directory or above will be merged together
     (in that order) using the read method of ConfigParser.
     """
-    config = RawConfigParser()
+    config = configparser.RawConfigParser()
 
     cli_conf = options.config
 
@@ -2105,7 +2105,16 @@ def read_config(options, args, arglist, parser):
 
     parent = tail = args and os.path.abspath(os.path.commonprefix(args))
     while tail:
-        if config.read(os.path.join(parent, fn) for fn in PROJECT_CONFIG):
+        try:
+            files_read = config.read(os.path.join(parent, fn)
+                                     for fn in PROJECT_CONFIG)
+        except configparser.ParsingError:
+            if options.verbose:
+                print('failed to read local configuration files due to a '
+                      'parsing error')
+            files_read = []
+
+        if files_read:
             local_dir = parent
             if options.verbose:
                 print('local configuration: in %s' % parent)
--- a/testsuite/test_shell.py
+++ b/testsuite/test_shell.py
@@ -11,11 +11,12 @@ class ShellTestCase(unittest.TestCase):
     """Test the usual CLI options and output."""
 
     def setUp(self):
+        self._RawConfigParser = pycodestyle.configparser.RawConfigParser
         self._saved_argv = sys.argv
         self._saved_stdout = sys.stdout
         self._saved_stderr = sys.stderr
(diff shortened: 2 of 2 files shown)


pycodestyle.py
```diff
@@ -2105,7 +2105,16 @@ def read_config(options, args, arglist, parser):
...
     parent = tail = args and os.path.abspath(os.path.commonprefix(args))
     while tail:
-        if config.read(os.path.join(parent, fn) for fn in PROJECT_CONFIG):
+        try:
+            files_read = config.read(os.path.join(parent, fn)
+                                     for fn in PROJECT_CONFIG)
+        except configparser.ParsingError:
+            if options.verbose:
+                print('failed to read local configuration files due to a '
+                      'parsing error')
```
jayvdb: Print the exception message?

jayvdb:
This can lead to breaking changes (inc. Upstream changes) not being noticed.

If the config is wrong, it should exit with a status code, otherwise this changes is a command line api breaking change.
And not raising an exception is a python api breaking change. Not good. imo the exception should be re-raised as-is, and then possibly caught elsewhere.

pycodestyle.py
```diff
@@ -2105,7 +2105,16 @@ def read_config(options, args, arglist, parser):
 
     parent = tail = args and os.path.abspath(os.path.commonprefix(args))
     while tail:
-        if config.read(os.path.join(parent, fn) for fn in PROJECT_CONFIG):
+        try:
+            files_read = config.read(os.path.join(parent, fn)
+                                     for fn in PROJECT_CONFIG)
+        except configparser.ParsingError:
+            if options.verbose:
+                print('failed to read local configuration files due to a '
```
IanLee1521: @sigmavirus24 -- It would be nice if we also printed out the list of files: `os.path.join(parent, fn) for fn in PROJECT_CONFIG` so that a user could narrow down which ones caused the ParsingError.

sigmavirus24: That works for me. I was going to include that initially but wasn't sure how much refactoring you would be comfortable with.

pycodestyle.py
sigmavirus24: I don't have strong opinions on that. It's not exactly the most useful error message and it will certainly make the verbose output less readable for most people.

sigmavirus24:
> If the config is wrong, it should exit with a status code, otherwise this changes is a command line api breaking change.

I disagree strongly. If a config is "broken" in the myriad of ways it can be broken, I don't think we should be breaking out early. The fact that we do now is due to us not handling an error case. Saying this is backwards incompatible is tantamount to saying that a web server once didn't handle an exception and returned a 500 server error and thus should continue to do so until the next "major" release.

> And not raising an exception is a python api breaking change. Not good. imo the exception should be re-raised as-is, and then possibly caught elsewhere.

You mean that people importing pycodestyle expect third-party exceptions to be raised instead of pycodestyle exceptions (which don't exist)? That makes little sense.

---

Most users will notice if their config file suddenly stops working when they run pycodestyle/flake8/whatever. If they then start using the verbose flags to look for why, they'll see this new message and investigate their config files.

As far as both pycodestyle and flake8 are concerned, the configuration files it looks for are non-essential to the execution of the program. If they were essential, than exiting immediately with a non-zero exit code would make perfect sense.

I also understand the concern that this is going to cause lots of problems but in reality we've had one such bug report in the *many* years of pycodestyle/pep8's existence.

As I see it right now, we can take two directions with this pull request:

- We can start loading the config files individually inside the while loop (so nested loops) and if one fails, we print out it's name and the fact that it could not be parsed when passed `-v`
- We can instead get the list of file names and print them out with the message instead

I'm okay with either. The former shouldn't cause much of a performance degradation for the average project because they shouldn't have a ton of setup.cfg/tox.ini files scattered through sub-directories and such.

I leave that decision to you, @IanLee1521

pycodestyle.py
ambv: Do you have an example of an non-useful error message? Any ideas how the error message could be improved?

sigmavirus24: @ambv it's been a while since I worked on this. I'd look at the example config that was probably provided in the issue.

PR #552, opened 2016-06-24
Simplify test running directions

FichteFoll:
The wording was poor, imo.

### DIFF
--- a/docs/developer.rst
+++ b/docs/developer.rst
@@ -104,8 +104,7 @@ Then be sure to pass the tests::
 
 When contributing to pycodestyle, please observe our `Code of Conduct`_.
 
-To run the Travis CI tests, you'll have to create a build the pycodestyle
-package using tox::
+To run the tests, use tox::
 
     $ pip install -r dev-requirements.txt
     $ tox


docs/developer.rst
```diff
@@ -104,8 +104,7 @@ Then be sure to pass the tests::
 
 When contributing to pycodestyle, please observe our `Code of Conduct`_.
 
-To run the Travis CI tests, you'll have to create a build the pycodestyle
-package using tox::
+To run the tests, use tox::
```
sigmavirus24: What about "To run the tests the same way Travis CI runs them, use tox::"

I think that captures the intent of the original awkward phrasing.

FichteFoll: To me, the intend is "run tests". Instead of running test "the way Travis CI does", CI should be running the tests as *we* do.

sigmavirus24: The core developers use tox. Not everyone will want to. That's the point of specifying that you're running them the way the CI runs them - if they get incongruous results, they can replicate that.

sigmavirus24:
You closed this by accident, yes?

FichteFoll:
No, it was on purpose.

Since you disagree with my fundamental opinion on why I made the change to remove "Travis CI" and I presented all my arguments already, I don't see a future for this PR. The change is not important enough by any means to warrant spending more time on it.

Issue #561, opened 2016-07-24
Choice type parameters raise AssertionError for config files


xZise:
When using Flake8 2.6.2 together with pycodestyle 2.0.0 and [zheller/flake8-quotes](https://github.com/zheller/flake8-quotes) 0.7.0 I get the following AssertionError if `quotes` or `inline-quotes` are defined as parameter:

```
$ flake8 setup.py 
Traceback (most recent call last):
  File "/home/xzise/.pyenv/versions/3.5.2/bin/flake8", line 11, in <module>
    sys.exit(main())
  File "/home/xzise/.pyenv/versions/3.5.2/lib/python3.5/site-packages/flake8/main.py", line 28, in main
    flake8_style = get_style_guide(parse_argv=True)
  File "/home/xzise/.pyenv/versions/3.5.2/lib/python3.5/site-packages/flake8/engine.py", line 234, in get_style_guide
    styleguide = StyleGuide(**kwargs)
  File "/home/xzise/.pyenv/versions/3.5.2/lib/python3.5/site-packages/flake8/engine.py", line 154, in __init__
    self._styleguide = kwargs.pop('styleguide', NoQAStyleGuide(**kwargs))
  File "/home/xzise/.pyenv/versions/3.5.2/lib/python3.5/site-packages/pycodestyle.py", line 1833, in __init__
    arglist, parse_argv, config_file, parser)
  File "/home/xzise/.pyenv/versions/3.5.2/lib/python3.5/site-packages/pycodestyle.py", line 2120, in process_options
    options = read_config(options, args, arglist, parser)
  File "/home/xzise/.pyenv/versions/3.5.2/lib/python3.5/site-packages/pycodestyle.py", line 2076, in read_config
    assert opt_type in ('store_true', 'store_false')
AssertionError
```

This is because the option's type is `choice` and that isn't actually handled. It should be handled the same as `string` as it is handled the same [by optparse](https://docs.python.org/2/library/optparse.html#standard-option-types).

So instead of `elif opt_type == 'string':` it could just do `elif opt_type in ('string', 'choice'):`. This will also verify that the choice is actually correct so no additional changes would be required.

PR #562, opened 2016-07-25
Stop checking for string option type

sigmavirus24:
There's no need to explicitly check for a string type when parsing the
configuration file. When it's neither an integer or a boolean, the only
value it can logically be is string-like.

Closes gh-561


IanLee1521:
Thanks @sigmavirus24 !

PR #568, opened 2016-08-05
fix: Cast `verbose` to int in continued_indentation() (fixes #567)

### LINKED ISSUE #567: `verbose` passed as a string when an int is expected
After trying to add the line `vebose = 2` to my setup.cfg for flake8, the `continued_indentation()` function errors with with a `TypeError`. It looks like the function was expecting `verbose` to be an integer, but the config parser reads it as a string.

As an easy fix/hack, you could cast `verbose` to an int at the start of the function. Depending on the number of other projects depending on this function, that might be easier than adjusting each config parser individually.

Traceback:

  File "/usr/lib/python3.5/multiprocessing/process.py", line 249, in _bootstrap
    self.run()
  File "/usr/lib/python3.5/multiprocessing/process.py", line 93, in run
    self._target(*self._args, **self._kwargs)
  File "/home/michael/.virtualenvs/markdoc-clone/lib/python3.5/site-packages/flake8/checker.py", line 46, in _run_checks_from_queue
    checker.run_checks(results_queue, statistics_queue)
  File "/home/michael/.virtualenvs/markdoc-clone/lib/python3.5/site-packages/flake8/checker.py", line 579, in run_checks
    self.process_tokens()
  File "/home/michael/.virtualenvs/markdoc-clone/lib/python3.5/site-packages/flake8/checker.py", line 563, in process_tokens
    self.handle_newline(token_type)
  File "/home/michael/.virtualenvs/markdoc-clone/lib/python3.5/site-packages/flake8/checker.py", line 606, in handle_newline
    self.run_logical_checks()
  File "/home/michael/.virtualenvs/markdoc-clone/lib/python3.5/site-packages/flake8/checker.py", line 515, in run_logical_checks
    for offset, text in results:
  File "/home/michael/.virtualenvs/markdoc-clone/lib/python3.5/site-packages/pycodestyle.py", line 460, in continued_indentation
    if verbose >= 3:
TypeError: unorderable types: str() >= int()

Michael-F-Bryan:
Added a simple `verbose = int(verbose)` to the top of the function. Afterwards I ran `tox` and all the tests still passed. 

I haven't written a test to make sure this case is covered, partly because I'm not sure how the internals of `pycodestyle` are structured, and also because it seems to be a problem with a fairly trivial solution.

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -443,6 +443,9 @@ def continued_indentation(logical_line, tokens, indent_level, hang_closing,
     if noqa or nrows == 1:
         return
 
+    # Make sure verbose is recieved as an integer
+    verbose = int(verbose)
+
     # indent_next tells us whether the next block is indented; assuming
     # that it is indented by 4 spaces, then we should not allow 4-space
     # indents on the final continuation line; in turn, some other


sigmavirus24:
This doesn't belong here. This belongs in Flake8. It parses the configuration and should be parsing this as an integer to pass to pycodestyle.

PR #580, opened 2016-10-24
Fix: os.walk() should have followlinks set to True.

jgeysens:
Fix: os.walk() should have followlinks set to True in order to also follow symbolic links when evaluating files.


jayvdb:
Why?
This would be a major change which could cause CI workers to descend paths that were previously excluded, and have non-compliant files causing build failures.
Add a config option to enable this new behaviour?

Issue #573, opened 2016-09-15
2.1 release


daspecster:
I was wondering if it would be possible to cut the 2.1 release?

The https://github.com/PyCQA/pycodestyle/milestone/13 milestone appears to be complete.

If there's anything else that should be done before the release, I'm happy to help.

IanLee1521:
Thanks for asking @daspecster ! Actually one thing that would be super helpful, and which I dropped the ball on is updating the changelog with everything that's happened since 2.0. Would that be something you would be willing to help with?

daspecster:
@IanLee1521, sure no problem!

IanLee1521:
@sigmavirus24 in particular, but anyone that wants to weigh in -- I was just looking over things for the 2.1 release, and I wanted to run something by you. One of the issues that was fixed was #530 which I believe may have had an unexpected side effect.

By changing the name of the program (https://github.com/PyCQA/pycodestyle/blob/master/pycodestyle.py#L2050) from `pep8` to `pycodestyle` this changes the config section name from `pep8` to `pycodestyle` which would break (ignore) anyone's config that was using the old name. In the long run, that is what we want. But as a backwards incompatible change, I think it's an issue to address.

I see two paths forward. First is to add logic that would handle either name for the section (around https://github.com/PyCQA/pycodestyle/blob/master/pycodestyle.py#L2148), but with a warning that the `pep8` section name is deprecated in preference to `pycodestyle`. Or second, keep the change as is, but note it as backwards incompatible and release the next version as 3.0 instead of 2.1. This would also allow for merging in https://github.com/PyCQA/pycodestyle/pull/433.

Any thoughts (from anyone) on these two solutions?

daspecster:
My vote would be for option 2 with a message that the '[pep8]' section was deprecated.

Supporting  the `['pycodestyle]` section is actually one of the changes that I'm looking for.
https://github.com/GoogleCloudPlatform/google-cloud-python/pull/2020

sigmavirus24:
I would opt for option 1. But option 2 is suitable for me too.

daspecster:
@sigmavirus24 @IanLee1521, WDYT about cutting a release? Are there any other issues I could help with to get to 2.1 or 3.0?

sigmavirus24:
@IanLee1521 I'm happy to help by cutting the release.

sigmavirus24:
@IanLee1521 I'm happy to help by cutting the release.

[...]

PR #591, opened 2016-11-03
Check for both [pep8] and [pycodestyle] config sections.

daspecster:
Towards https://github.com/PyCQA/pycodestyle/issues/573.

This is options 1 and 2 I think that were mentioned there.

Support for both `[pep8]` and `[pycodestyle]` in config/ini. There is also a warning generated mentioning the deprecation of the `[pep8]` section.

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -55,6 +55,7 @@ import re
 import sys
 import time
 import tokenize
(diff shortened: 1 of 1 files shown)


daspecster:
@IanLee1521 no problem!

PR #592, opened 2016-11-06
Warn about bare except clause

rogalski:
Solves #579

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -912,8 +912,10 @@ def module_imports_on_top_of_file(
     Okay: # this is a comment\nimport os
     Okay: '''this is a module docstring'''\nimport os
     Okay: r'''this is a module docstring'''\nimport os
-    Okay: try:\n    import x\nexcept:\n    pass\nelse:\n    pass\nimport y
-    Okay: try:\n    import x\nexcept:\n    pass\nfinally:\n    pass\nimport y
+    Okay:
+    try:\n\timport x\nexcept ImportError:\n\tpass\nelse:\n\tpass\nimport y
+    Okay:
+    try:\n\timport x\nexcept ImportError:\n\tpass\nfinally:\n\tpass\nimport y
     E402: a=1\nimport os
     E402: 'One string'\n"Two string"\nimport os
     E402: a=1\nfrom sys import x
@@ -1177,6 +1179,22 @@ def comparison_type(logical_line, noqa):
         if inst and isidentifier(inst) and inst not in SINGLETONS:
             return  # Allow comparison for types which are not obvious
         yield match.start(), "E721 do not compare types, use 'isinstance()'"
+
+
+def bare_except(logical_line, noqa):
+    r"""When catching exceptions, mention specific exceptions whenever possible.
+
+    Okay: except Exception:
+    Okay: except BaseException:
+    E722: except:
+    """
+    if noqa:
+        return
+
+    regex = re.compile(r"except\s*:")
+    match = regex.match(logical_line)
+    if match:
+        yield match.start(), "E722 do not use bare except'"
 
 
 def ambiguous_identifier(logical_line, tokens):
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -28,6 +28,7 @@ Changes:
 * Report E741 on 'global' and 'nonlocal' statements, as well as prohibited
   single-letter variables.
 * Deprecated use of `[pep8]` section name in favor of `[pycodestyle]`; #591
+* Report E722 when bare except clause is used; #579
 
 Bugs:
 
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -361,6 +361,8 @@ This is the current list of error and warning codes:
 +------------+----------------------------------------------------------------------+
 | E721 (^)   | do not compare types, use 'isinstance()'                             |
 +------------+----------------------------------------------------------------------+
+| E722       | do not use bare except, specify exception instead                    |
++------------+----------------------------------------------------------------------+
 | E731       | do not assign a lambda expression, use a def                         |
 +------------+----------------------------------------------------------------------+
 | E741       | do not use variables named 'l', 'O', or 'I'                          |
--- a/testsuite/E30.py
+++ b/testsuite/E30.py
@@ -113,7 +113,7 @@ def a():
 
 try:
     a()
-except:
+except Exception:
     pass
 #: E305:5:1
 def a():
--- a/testsuite/E40.py
+++ b/testsuite/E40.py
@@ -18,7 +18,7 @@ import foo
 #: Okay
 try:
     import foo
-except:
+except ImportError:
     pass
 else:
     print('imported foo')
--- a/testsuite/E72.py
+++ b/testsuite/E72.py
@@ -49,3 +49,31 @@ if isinstance(res, types.MethodType):
     pass
 if type(a) != type(b) or type(a) == type(ccc):
     pass
+#: E722
+try:
+    pass
+except:
+    pass
+#: E722
+try:
+    pass
+except Exception:
+    pass
+except:
+    pass
+#: E722 E203 E271
+try:
+    pass
+except  :
+    pass
+#: Okay
+fake_code = """"
+try:
+    do_something()
+except:
+    pass
+"""
+try:
+    pass
+except Exception:
+    pass


spaceone:
Does this also blame the following?
```
try:
    foo()
except:
    self.log('foo failed')
    raise
```
It would be nice if this detects re-raising and don't blame then.

spaceone:
It would also be nice if it doesn't blame the following code:
```
try:
    foo()
except (SystemExit, KeyboardInterrupt, SyntaxError):
    raise
except:
    bar()
```

rogalski:
I'm quite sure that won't be possible in current pycodestyle philosophy, where tool is lightweight and no AST is built for purpose of yielding warnings. Hopefully one of maintainers can confirm.

sigmavirus24:
@spaceone the correct forms of both of those constructs is to use `except Exception` unless you're writing Python that is meant to run on Python 2.4 or earlier.

sigmavirus24:
New checks are only added in minor releases, so I've marked this for the 2.2.0 release (or the 2.2 milestone). A quick look tells me this looks good @rogalski. Thanks!

The-Compiler:
I think there are good reasons to use a bare `except:` and re-raising - imagine something like this with a library not providing context managers:

```python
transaction = ...
try:
    ...
except:
    transaction.rollback()
    raise
```

Here `except:` is the right thing to use, as you'd want to do the rollback even when there's e.g. a `KeyboardInterrupt`.

You could use `except BaseException:` on Python 3, but I feel like that is pretty subjective. Even pep8 (the document) says:

> A good rule of thumb is to limit use of bare 'except' clauses to two cases:
>
> - If the exception handler will be printing out or logging the traceback; at least the user will be aware that an error has occurred.
> **- If the code needs to do some cleanup work, but then lets the exception propagate upwards with raise . try...finally can be a better way to handle this case.**

(my emphasis).

IMHO, this kind of check is too aggressive for (what I preceive to be) pycodestyle's philosophy, and better left to more aggressive and sophisticated checkers like pylint (which already does this check, and doesn't emit the message when re-raising).

sigmavirus24:
> IMHO, this kind of check is too aggressive for (what I preceive to be) pycodestyle's philosophy

I don't agree (but I also don't entirely disagree). Here's why: Those two cases are two very special cases and what the part you're not quoting is specifically warning about is what I expect this check would cat 90+% of the time.

For the cases you're discussing, I expect those to be *rare* enough and advanced enough that the developer would know when to use `# noqa` to silence that particular item. So, from that perspective I think this is perfectly acceptable.

That said, I'm also not convinced that this won't cause a glut of headaches for folks using bare excepts correctly today. Perhaps a decent compromise would be to add this but also include it in the default ignore list so that people who do not want this style used (except in very rare cases) can turn it on without us causing anyone else significant pain and suffering.

spaceone:
This case is not barely used!

rogalski:
Are any major open source libraries using bare excepts in this manner?

On the other hand, I honestly feel like 99% of bare excepts implemented in Python code everyday are done because of sloppiness or unawareness and not because it's a correct approach for solving given problem.

Having it as an opt-in seems like a reasonable compromise, but rejecting it entirely is also fine. I guess maintainers have to make a decision.

sigmavirus24:
> Are any major open source libraries using bare excepts in this manner?

The largest body of open source Python I've seen (OpenStack) explicitly avoids bare except clauses like this and tends to use either `Exception` or `BaseException`. In fact I'd be intrigued to see if this would catch any bare except clauses in those projects (my guess is no).

>  I honestly feel like 99% of bare excepts implemented in Python code everyday are done because of sloppiness or unawareness and not because it's a correct approach for solving given problem.

I think sloppiness isn't the correct thing here. I think it's more of "Person X over in Department Y needs this to be working by the end of the day and I can't handle every exception case right now. I'll fix this later." In other words, they didn't have the time to do it. I would hesitate to call that anything other than efficiency.

> I guess maintainers have to make a decision.

As a maintainer, I'd like this check included. I'll leave the final call (as to whether this is on by default or ignored by default) to @IanLee1521 though.

rogalski:
> Person X over in Department Y needs this to be working by the end of the day and I can't handle every exception case right now. I'll fix this later

The thing is - correct solution for this case in Python is to use `except Exception:`. You may call it _an efficiency_, but for me it's unawareness of Python internals...

pycodestyle.py
```diff
@@ -1177,6 +1179,24 @@ def comparison_type(logical_line, noqa):
...
+    E722: except BaseException:
+    """
+    if noqa:
+        return
+
+    regex = re.compile(r"except(?:\s*)([A-z_]+[A-z0-9_])?(?:\s*):")
+    match = regex.match(logical_line)
+    if match:
+        exception_name = match.group(1)
+        if exception_name in (None, 'BaseException'):
```
rogalski: Maybe `except BaseException:` should be allowed.

spaceone:
`except:` and `except BaseException:` is the same.
https://docs.python.org/3/library/exceptions.html#exception-hierarchy

sigmavirus24:
> You may call it an efficiency, but for me it's unawareness of Python internals...

A lot of the people I work with might call it both. ;)

spaceone:
>     You may call it an efficiency, but for me it's unawareness of
>     Python internals...
>
> A lot of the people I work with might call it both. ;)
>
And what about the people who know about the python internals? I have
valid reasons to use "except:" and I don't want to waste my code with
different '# noqa'-like statements for every style check tool which
exists. (also '#noqa is unspecific about what shouldn't be QAed).

I don't want this code to be false-positive detected and yes I want to
catch also KeyboardInterrupt, SystemExit and GeneratorExit.
```
  2429 »   »   try:
  2430 »   »   »   super(Object, self)._move(newdn, modify_childs)
  2431 »   »   except:
  2432 »   »   »   # self couldn't be moved
  2433 »   »   »   # move back subelements and reraise
  2434 »   »   »   self.move_subelements(tmpdn, olddn, subelements)
  2435 »   »   »   raise

```

sigmavirus24:
@spaceone it seems you did not read the entire thread. Please do that before you continue with these aggressive comments. If you are confused about what we're discussing, feel free to ask questions and I'll be happy to clarify things. I think there's some agreement that this should start out (or permanently exist) in the `DEFAULT_IGNORE` list, so your comments are not constructive.

spaceone:
Sorry for the agressiveness. I of course read and am involved in the whole thread since the beginning.
And it seems my suggestion to ignore re-raised exception blocks will still cause this error (even if the check is disabled by default, or in some DEFAULT_IGNORE list whatever this is).

pycodestyle.py
spaceone: The code also doesn't detect `except (ValueError, BaseException):`

sigmavirus24:
@spaceone Right, it was becoming apparent that you didn't know what `DEFAULT_IGNORE` was. This is a collection of error codes that pycodestyle does *not* check for in a default configuration (i.e., you don't specify your own `--ignore`) In other words, adding this there means it won't be checked for unless people select it or remove it from their ignore list.

rogalski:
Is there a reasonable consensus on what should be covered by this checker?

I want to clean up my dangling pull requests - either rebase and finish implementation or throw them away.

ambv:
[Bugbear](https://github.com/PyCQA/flake8-bugbear) started out as a project because of bare `except:` clauses in codebases I work with. See the error message for B001.

What I learned with experience and was confirmed by talking with Raymond and Guido is that "except:" is often used because the user didn't know any better. @spaceone's use case for catching BaseException is reasonable, there's cases we really need that. In this case it's better to be explicit is write "except BaseException". This is what Guido himself is doing when coding, see example: https://hg.python.org/cpython/file/3.6/Lib/asyncio/selector_events.py#l875

My advice would be to warn against using bare `except:` and allow `except BaseException:`.

rogalski:
After re-reading whole pull request comment history, I really feel like warning strictly about `bare except:` and allowing all other versions will be a most reasonable choice, especially because _explicit is better than implicit_. I'll revisit this change and push new revision soon.

@ambv do you know if BDFL have ever made public statement on this matter? I think that having it 'officially' (instead of _confirmed by Guido in (private, I assume) talks_) would be a significant factor in determining if PR should be accepted or rejected.

sigmavirus24:
I'm rather comfortable with this. Thanks for submitting this @rogalski and thanks for the input @spaceone and @ambv :)

PR #593, opened 2016-11-06
Fix E305 regression in 2.1.0 due to fix for #400

markpeek:
The fix used in #400 did not account for usage of "def" or "class" at the beginning of variables such as "default" or "classify". This PR looks for a space at the end of these keywords.


sigmavirus24:
Thanks @markpeek!

PR #595, opened 2016-11-07
Add E306 to docs

spaceone:


spaceone:
You have ," found 0" also in other messages. I find it good that it is there because this is what actually is printed out.

spaceone:
Flake8 is even better in this as it prints '...' or "name" in italic for variable things.

spaceone:
See http://flake8.pycqa.org/en/latest/user/error-codes.html

docs/intro.rst
```diff
@@ -314,17 +314,17 @@ This is the current list of error and warning codes:
 +------------+----------------------------------------------------------------------+
 | **E3**     | *Blank line*                                                         |
 +------------+----------------------------------------------------------------------+
-| E301       | expected 1 blank line, found 0                                       |
+| E301       | expected 1 blank line, found …                                       |
```
sigmavirus24: This now exceeds the stated purpose of this pull request and will break downstream redistributors who don't have an internationalized locale on their builders and CI infrastructure. Please do not use unicode here.

spaceone: So, what do you want? just removal of ", found 0"? Where? in E301 or in all E*? Or replacing with "..." (ascii)?

sigmavirus24: Let's just stick to adding the new message. Then let's update these in a separate pull request.

spaceone:
https://github.com/PyCQA/pycodestyle/pull/595/commits/349deeae7f03a77954189553f5eace8fe0e980ec

sigmavirus24:
I'll let @IanLee1521 weigh in, but this looks fine to me.

IanLee1521:
Thanks! @spaceone !

PR #610, opened 2017-01-16
 added function doc to blank_lines() for E306

khairihafsham:
without it `register_check()` using ERRORCODE_REGEX will not register E306 into error code list. This in effect will make `pycodestyle --select=E306` to not work.


sigmavirus24:
Thanks @khairihafsham!

PR #611, opened 2017-01-23
Report E704 for async def as well

asottile:


testsuite/E70.py
```diff
@@ -12,6 +12,8 @@
 del a[:]; a.append(42);
 #: E704:1:1
 def f(x): return 2
+#: E704:1:1
+async def f(x): return 2
```
spaceone: Does it detect two spaces?
```
#: E704:1:1
+async  def f(x): return 2
```

asottile: indeed it does not, I can extend this further

asottile: I've fixed this case as well as two others where 'async  def' was not being detected

sigmavirus24: In my opinion 2 spaces is not something that should be caught. That should trigger a different warning, probably along the lines of the warnings that are triggered when there are too many spaces around a `=`.

asottile: It also triggers that, but now triggers both

asottile: I think it's beneficial to trigger both warnings -- it reduces the number of steps to finish refactoring some bad code.

With this patch it immediately shows both warnings, but without this patch you would first see E272, fix it, and then see E704.

asottile: Or if someone (for some reason) disabled E272 but not E704

sigmavirus24: Good point. That works for me. :)

PR #612, opened 2017-01-23
Don't report E701 for variable annotations

### LINKED ISSUE #575: Support syntax for variable annotation in Python 3.6
[PEP 526](https://www.python.org/dev/peps/pep-0526/) introduced the syntax for variable annotations in Python 3.6. At the moment pycodestyle.py warns of every such annotation taking them for "E701 multiple statements on one line (colon)":

``` python
from typing import ClassVar, List

CONST: int = 42  # E701


class Class:
    cls_var: ClassVar[str]  # E701

    def m(self):
        xs: List[int] = []  # E701
```

```
$ python3.6 pycodestyle.py variable_annotations.py 
variable_annotations.py:3:6: E701 multiple statements on one line (colon)
variable_annotations.py:7:12: E701 multiple statements on one line (colon)
variable_annotations.py:10:11: E701 multiple statements on one line (colon)
```

asottile:
This is based on #611 to avoid a conflict.  I can rebase / reorder if requested.

This resolves #575

### DIFF
--- a/testsuite/python3.py
+++ b/testsuite/python3.py
@@ -1,6 +1,18 @@
 #!/usr/bin/env python3
+from typing import ClassVar, List
 
 
 # Annotated function (Issue #29)
 def foo(x: int) -> int:
     return x + 1
+
+
+# Annotated variables #575
+CONST: int = 42
+
+
+class Class:
+    cls_var: ClassVar[str]
+
+    def m(self):
+        xs: List[int] = []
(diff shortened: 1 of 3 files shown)


sigmavirus24:
@asottile please rebase.

asottile:
rebased!  Should be ready for review :)

testsuite/python3.py
```diff
@@ -1,6 +1,18 @@
 #!/usr/bin/env python3
+from typing import ClassVar, List
```
sigmavirus24: So we support several versions of Python 3 including those without this module. I wonder if it wouldn't hurt to have a separate `python36.py` file that uses the typing module and annotated variables, etc.

Thoughts?

asottile: I probably wouldn't bother?  The file isn't actually being run

sigmavirus24:
Something's up with Travis-CI. Closing and reopening

asottile:
<3 Looking forward to the python3.6 release!!!

Issue #617, opened 2017-01-30
E302 False-positive for variable names starting with `def`


willhardy:
Given the following code:

```
defaults = {}
defaults.update()
```

A false positive is reported: `E302 expected 2 blank lines, found 0`.

Changing of the variable name to something that doesn't start with `def` avoids the false positive. This is only triggered in 2.3.0, not 2.2.0.

gpaul:
We're affected by this, too.

bronsen:
As are we.

sigmavirus24:
I submitted #618 to fix this @willhardy @gpaul @bronsen. If y'all who are affected by this test that out to double-check my work, I can get a patch release out this afternoon/evening.

gpaul:
Just checked. I can confirm your patch fixes this issue.

sigmavirus24:
The fix for this was merged and released in 2.3.1.

PR #618, opened 2017-01-30
Fix regression in E302 detection

sigmavirus24:
This was similar to an issue we had with E306 in which we were only
checking that a line started with "def" or "class" which catches global
vars like "defaults" or "classification".

Closes gh-617

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -122,7 +122,7 @@ OPERATOR_REGEX = re.compile(r'(?:[^,\s])(\s*)(?:[-+*/|!<=>%&^]+)(\s*)')
 LAMBDA_REGEX = re.compile(r'\blambda\b')
 HUNK_REGEX = re.compile(r'^@@ -\d+(?:,\d+)? \+(\d+)(?:,(\d+))? @@.*$')
 STARTSWITH_DEF_REGEX = re.compile(r'^(async\s+def|def)')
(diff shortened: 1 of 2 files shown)


pycodestyle.py
```diff
@@ -122,7 +122,7 @@
 LAMBDA_REGEX = re.compile(r'\blambda\b')
 HUNK_REGEX = re.compile(r'^@@ -\d+(?:,\d+)? \+(\d+)(?:,(\d+))? @@.*$')
 STARTSWITH_DEF_REGEX = re.compile(r'^(async\s+def|def)')
```
spaceone: what about this one?

sigmavirus24: Not all regular expressions named similarly need to be touched all at once. This one is actually fine as is.

pycodestyle.py
```diff
@@ -122,7 +122,7 @@
...
-STARTSWITH_TOP_LEVEL_REGEX = re.compile(r'^(async\s+def|def|class|@)')
+STARTSWITH_TOP_LEVEL_REGEX = re.compile(r'^(async\s+def |def |class |@)')
```
spaceone: I would use "\s" to also detect tabs!

sigmavirus24: Why are you exclaiming this?!

sigmavirus24: Good idea! Updated!

PR #625, opened 2017-02-24
Python 3.6 invalid escape sequence deprecation fix

scop:


scop:
Added a link to the Python 3.6 release notes section that has the explanation. The effect is the same as with deprecated features in general, a warning will be emitted if deprecation warnings are enabled. I don't think that needs to be explained in the commit message.

sigmavirus24:
> I don't think that needs to be explained in the commit message.

Links break. At the very least, please quote the relevant section

```
A backslash-character pair that is not a valid escape sequence now generates a DeprecationWarning. Although this will eventually become a SyntaxError, that will not be for several Python releases.
```

scop:
(Just a ping to note that the requested changes have been done here a while ago...)

sigmavirus24:
Sorry about that @scop. Thanks for addressing that! :cake:

PR #630, opened 2017-03-15
Correctly report E501 when the first line of a docstring is too long

### LINKED ISSUE #622: max line length not enforced for long first line docstrings
Somewhat new to python tooling and pycodestyle, but I believe this should trigger the (default) max line length E501:

```
def foo():
    """Lorem ipsum dolor sit amet, consectetur adipiscing elit. Duis pulvinar vitae

    Lorem ipsum dolor sit amet, consectetur adipiscing elit.
    """
```

The first docstring line is 83 characters (> 79). I believe pycodestyle may be looking at the tokenized docstring, which doesn't include the leading whitespace in front of the opening """ in the physical line.

asottile:
Resolves #622

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -1199,7 +1199,7 @@ def comparison_type(logical_line, noqa):
 
 
 def bare_except(logical_line, noqa):
-    r"""When catching exceptions, mention specific exceptions whenever possible.
+    r"""When catching exceptions, mention specific exceptions when possible.
 
     Okay: except Exception:
     Okay: except BaseException:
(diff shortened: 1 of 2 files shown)


pycodestyle.py
```diff
@@ -1199,7 +1199,7 @@ def comparison_type(logical_line, noqa):
 
 
 def bare_except(logical_line, noqa):
-    r"""When catching exceptions, mention specific exceptions whenever possible.
+    r"""When catching exceptions, mention specific exceptions when possible.
```
asottile: this line was too long, I tried shortening it without changing meaning

asottile:
@sigmavirus24  thoughts on this?

asottile:
@sigmavirus24 bump again :)

sigmavirus24:
Sorry for the delay @asottile. I've been quite busy for a while now.

PR #632, opened 2017-03-15
Option for detecting hanging indents

SamriddhiJain:


SamriddhiJain:
I want this error code to be ignored by default so that it doesn't interfere with the existing tests as it is doing now in the travis builds. So far I have added the code in `default_ignore` list in `pycodestyle.py, setup.cfg, test_api.py`, but with no luck. Is there any documentation which I can follow for this?

SamriddhiJain:
@sigmavirus24 Hi, I'm trying to add this test for hanging indents and would like to make it ignore by default. I followed this similar commit of yours for W503, [1](https://github.com/sigmavirus24/pep8/commit/27d638454ad34fca4c2d1b1f9ad40477b6c781bd), but still the tests are running. As I'm not able to find any documentation on how to do this, can you please suggest what am I missing to make this test ignore by default.

menshikh-iv:
Hi @SamriddhiJain, we are very interested in this PR, what's a status here?

sigmavirus24:
@menshikh-iv this is blocked on my help/review. I'm hopefully going to get around to this in the next month

PR #634, opened 2017-03-23
Support variable annotation when variable start by a keyword

rach:
Support case when the variable name start by a keyword, eg:
```
class Test:
    formula: str = None # E701
```

This  PR ensure that we do the codestyle check only for line which start by the keyword followed by a space.

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -121,10 +121,10 @@ KEYWORD_REGEX = re.compile(r'(\s*)\b(?:%s)\b(\s*)' % r'|'.join(KEYWORDS))
 OPERATOR_REGEX = re.compile(r'(?:[^,\s])(\s*)(?:[-+*/|!<=>%&^]+)(\s*)')
 LAMBDA_REGEX = re.compile(r'\blambda\b')
 HUNK_REGEX = re.compile(r'^@@ -\d+(?:,\d+)? \+(\d+)(?:,(\d+))? @@.*$')
-STARTSWITH_DEF_REGEX = re.compile(r'^(async\s+def|def)')
+STARTSWITH_DEF_REGEX = re.compile(r'^(async\b+def|def)')
 STARTSWITH_TOP_LEVEL_REGEX = re.compile(r'^(async\s+def\s+|def\s+|class\s+|@)')
 STARTSWITH_INDENT_STATEMENT_REGEX = re.compile(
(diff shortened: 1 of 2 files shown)


pycodestyle.py
```diff
@@ -124,7 +124,7 @@
 STARTSWITH_DEF_REGEX = re.compile(r'^(async\s+def|def)')
 STARTSWITH_TOP_LEVEL_REGEX = re.compile(r'^(async\s+def\s+|def\s+|class\s+|@)')
 STARTSWITH_INDENT_STATEMENT_REGEX = re.compile(
-    r'^\s*({0})'.format('|'.join(s.replace(' ', '\s+') for s in (
+    r'^\s*({0})\s'.format('|'.join(s.replace(' ', '\s+') for s in (
```
eddieantonio: Isn't `\b` more appropriate here? For example, this is valid Python 3 (and should be flagged by this rule), but is unsupported by your regex:

```python
while+False: ...
```

pycodestyle.py
```diff
@@ -124,7 +124,7 @@
 STARTSWITH_DEF_REGEX = re.compile(r'^(async\s+def|def)')
```
eddieantonio: You will also have to add `\b` here, as lines starting with `async? def` are special-cased in `def compound_statements(logical_line):`

testsuite/python3.py
```diff
@@ -13,6 +13,16 @@ def foo(x: int) -> int:
...
+    for_var: ClassVar[str]
+    while_var: ClassVar[str]
+    def_var: ClassVar[str]
+    if_var: ClassVar[str]
+    elif_var: ClassVar[str]
+    else_var: ClassVar[str]
+    try_var: ClassVar[str]
+    except_var: ClassVar[str]
+    finally_var: ClassVar[str]
+    with_var: ClassVar[str]
```
sigmavirus24: Might as well be *paranoid* and add camelCased versions of each of these while we're here.

pycodestyle.py
sigmavirus24: What value is there in allowing people to use that syntax?

eddieantonio: Haha, that's a good point! Though might as well not break this rule, and let that abomination be caught by E2xx rule.

sigmavirus24:
Subsumed by #640

sigmavirus24:
Thanks for contributing @rach!

PR #640, opened 2017-04-15
Fix variable annotations

eddieantonio:
This is an extension of #634, but with passing tests.


eddieantonio:
Bumping this PR. I haven't been able to write things like 

```python
forward_passes: List[Pass]
```
without failing the style check for a while!

sigmavirus24:
Thanks @eddieantonio !

pjdelport:
Can we please have a release for this fix?

PR #642, opened 2017-04-26
Add --no-visual flag to disallow visual indentation

bwhmather:
An proposal for allowing projects to enforce hanging indentation.

Adds a flag which disallows code that is indented to align with the opening of the containing brackets:
``` python
def function(argument_1,
            argument_2,
            argument_3):
    pass
```
It instead requires that the code be indented to a multiple of the indentation size:
``` python
def function(
    argument_1,
    argument_2,
):
    pass
```
This has a similar goal to #632 but is implemented by disabling the cases that accept visual indents, rather than accepting them initially but raising a new error type later.
Should result in cleaner error messages, and does not require a separate pass through the code.


bwhmather:
Hi - Is there any interest in this?

Issue #643, opened 2017-05-03
Slow treatment of very long lines


JulienPalard:
Hi,

Did not took the time to fully investigate, but running ``pycodestyle --exclude=.git --statistics --count .`` on a clone of ``github.com/PmagPy/PmagPy`` looks to never end.

Tested on commit 74a65794 of PmagPy and pycodestyle 2.2.0.

JulienPalard:
OK I'm trying to diagnose this one now, first hint, the culpit is: https://github.com/PmagPy/PmagPy/blob/master/pmagpy/coefficients.py so probably not an infinite loop, "just" a very long execution time (like more than 8 hours on my laptop if I remember well)

JulienPalard:
After running cProfile, it looks like 37s out of 42.8s are consumed by `noqa`.

The "problem" is `noqa` is called for every token of a line, so for a very long line with a lot of tokens it gets called a lots of time with the whole long line. On a 148KB files I measured more than 3.2GB of strings given to noqa (killed it before the end).

A simple fix may be to use a lru_cache, which sadly isn't available in 2.7 and I don't want to reimplement a lru_cache (errors = more code²) nor copy the cpython one (~100 more lines) I propse here to cache only if lru_cache is available, else just not fix this issue.

On a 148KB file with very long lines, some run times:
- Without lru_cache on noqa: 40.954s, 40.614s, 41.143s
- With lru_cache on noqa: 4.971s, 5.089s, 4.935s

PR #644, opened 2017-05-04
Fix issue #643: Optimize noqa() with an lru_cache for Python 3.2+.

JulienPalard:
Fixes https://github.com/PyCQA/pycodestyle/issues/643 for Python 3.2+.

Had to catch a "No signature found for builtin <built-in method
search of _sre.SRE_Pattern object at 0x...>" in 3.4:

In python3.4 the search was not detected as a function, now that it's
wrapped in an lru_cache it is, yet it still has no signature (it has
in 3.5+).

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -58,6 +58,16 @@ import tokenize
 import warnings
 import bisect
 
+try:
+    from functools import lru_cache
+except ImportError:
+    def lru_cache(maxsize=128):  # noqa as it's a fake implementation.
+        """Does not really need a real a lru_cache, it's just optimization, so
+        let's just do nothing here. Python 3.2+ will just get better
+        performances, time to upgrade?
+        """
+        return lambda function: function
+
 from fnmatch import fnmatch
 from optparse import OptionParser
 
@@ -1410,7 +1420,7 @@ else:
         """Read the value from stdin."""
         return TextIOWrapper(sys.stdin.buffer, errors='ignore').read()
 
-noqa = re.compile(r'# no(?:qa|pep8)\b', re.I).search
+noqa = lru_cache(512)(re.compile(r'# no(?:qa|pep8)\b', re.I).search)
 
 
 def expand_indent(line):


pycodestyle.py
```diff
@@ -56,6 +56,15 @@
 import time
 import tokenize
 import warnings
+try:
+    from functools import lru_cache
+except ImportError:
+    def lru_cache(maxsize=128):  # noqa as it's a fake implementation.
```
sigmavirus24: What versions of Python are we guarding against here? I think the minimum version we support now is 3.3.

pycodestyle.py
```diff
@@ -1476,10 +1485,15 @@ def _is_eol_token(token, _eol_token=_is_eol_token):
...
-                for parameter
-                in inspect.signature(function).parameters.values()
-                if parameter.kind == parameter.POSITIONAL_OR_KEYWORD]
+        try:
+            return [parameter.name
+                    for parameter
+                    in inspect.signature(function).parameters.values()
+                    if parameter.kind == parameter.POSITIONAL_OR_KEYWORD]
+        except ValueError:
+            """Catches "No signature found for builtin <built-in method
```
sigmavirus24: We merged the other code that doesn't rely on scanning the definitions in the module to find checks. Is this still necessary now? You'll need to rebase so I think this is probably unnecessary at this point.

pycodestyle.py
JulienPalard: According to your [tox.ini](https://github.com/PyCQA/pycodestyle/blob/master/tox.ini), it's 2.6. If the minimum supported version is really 3.3 now, it should probably be dropped from tox.ini?

pycodestyle.py
JulienPalard: It's in fact no longer needed since the merge (kina makes me happy ^^). Rebased on master.

pycodestyle.py
sigmavirus24: Sorry, I misspoke. The minimum supported Python 3 version is 3.3. I'm not sure if @IanLee1521 wants to continue to support Python 2.6.

JulienPalard: As long as `tox.ini` tells 2.6 and 2.7, I'll keep this try, it can easily be dropped later with the drop of 2.x support. I bet some other compatibility things will drop too in any way.

PR #648, opened 2017-05-06
Use bisect instead of iterating over every offsets.

JulienPalard:
As the offsets looks already sorted (which looks logical so I assumed
they always are), using a bisection if faster than iterating over all
of them.

On a "worst case scenario" I encontered, I got nice enhancement with this patch:

$ time python3 ./pycodestyle.py ~/Downloads/PmagPy/PmagPy/coefficients-552K.py > /dev/null

real 1m16.405s

$ time python3 ./pycodestyle.py ~/Downloads/PmagPy/PmagPy/coefficients-552K.py > /dev/null

real 0m3.318s

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -56,6 +56,7 @@ import sys
 import time
 import tokenize
 import warnings
+import bisect
 
 from fnmatch import fnmatch
 from optparse import OptionParser
@@ -1664,10 +1665,10 @@ class Checker(object):
         """Build a line from tokens and run all logical checks on it."""
         self.report.increment_logical_line()
         mapping = self.build_tokens_line()
-
         if not mapping:
             return
 
+        mapping_offsets = [offset for offset, _ in mapping]
         (start_row, start_col) = mapping[0][1]
         start_line = self.lines[start_row - 1]
         self.indent_level = expand_indent(start_line[:start_col])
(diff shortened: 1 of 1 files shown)


pycodestyle.py
```diff
@@ -1645,10 +1646,10 @@ def check_logical(self):
         """Build a line from tokens and run all logical checks on it."""
         self.report.increment_logical_line()
         mapping = self.build_tokens_line()
-
         if not mapping:
             return
 
+        mapping_offsets = [offset for offset, pos in mapping]
```
sigmavirus24: There's no need to bind `pos` here. This should more simply be

```py
mapping_offsets = [offset for offset, _ in mapping]
```

pycodestyle.py
```diff
@@ -1662,9 +1663,8 @@ def check_logical(self):
             self.init_checker_state(name, argument_names)
             for offset, text in self.run_check(check, argument_names) or ():
                 if not isinstance(offset, tuple):
-                    for token_offset, pos in mapping:
-                        if offset <= token_offset:
-                            break
+                    token_offset, pos = mapping[bisect.bisect_left(
```
sigmavirus24: I'd prefer it if you added a comment here explaining the benefit of `bisect_left` so this doesn't get replaced later on by a well-meaning person.

pycodestyle.py
JulienPalard: fair enough

pycodestyle.py
JulienPalard: done.

JulienPalard:
Fixed your thoughts and rebased on master.

sigmavirus24:
FWIW, thank you so much for including everything in the commit message. It means that if this project survives long enough for GitHub to go the way of Sourceforge that we'll still have an understanding of why we did this. 🎉 🍰

PR #651, opened 2017-05-30
Added .gitattributes to prevent auto-munging E90.py line endings

kjcole:
I have ~/.gitattributes that normalizes line endings for various files. However, E90.py intentionally contains a line with a "broken" weird line ending. I had a devil of a time figuring out why git insisted there was a differnce between my repo and the upstream master, even after I explicitly cloned your original until I realized git was auto-munging the line ending.

(Even if you don't take this pull request, you may want to add a .gitattributes to prevent such fixing.)


sigmavirus24:
Thanks @kjcole! I think this is worth having to avoid other folks who are less familiar with git running headlong into this. 🎉 🍰

PR #654, opened 2017-06-03
Document all supported Python versions as trove classifiers

jdufresne:
I frequently use PyPI trove classifiers to check if a 3rd party package
is usable by my projects. Documenting the supported versions makes this
much easier for potential library users.


sigmavirus24:
Thanks!

PR #657, opened 2017-06-15
Make tested code valid Python code

SylvainDe:
Playing with AST checks can be very tedious
because of the amount of "incorrect" Python
code which raises E901 (or fail differently)
when AST checks are enabled.

An easy solution is to fix the code but this
is not always applicable.

I've only performed the trivial changes.


pycodestyle.py
```diff
@@ -871,17 +871,15 @@ def whitespace_around_named_parameter_equals(logical_line, tokens):
     Don't use spaces around the '=' sign when used to indicate a
     keyword argument or a default parameter value.
 
-    Okay: def complex(real, imag=0.0):
-    Okay: return magic(r=real, i=imag)
+    Okay: def complex(real, imag=0.0):\n    return magic(r=real, i=imag)
```
sigmavirus24: These were two separate checks for two separate test cases. Collapsing them may introduce bugs

SylvainDe: The whole thing is still tested, isn't it ? Or I can write:

    def complex(real, imag=0.0): pass

and

    magic(r=real, i=imag)

testsuite/E11.py
```diff
@@ -1,13 +1,15 @@
 #: E111
 if x > 2:
-  print x
+  print(x)
```
sigmavirus24: This means we'll loose coverage for the print statement then, yes?

testsuite/E12.py
```diff
@@ -1,22 +1,22 @@
 #: E121
-print "E121", (
```
sigmavirus24: Same concern as above

pycodestyle.py
sigmavirus24: I would rather retain two separate tests cases.

SylvainDe: I am suggesting 2 separate test cases, each of them corresponding to a piece of valid Python code.

testsuite/E11.py
SylvainDe: I am not sure I fully understand what you mean. The E11 errors does not seem to be related to print and/or parentheses in any way.

testsuite/E12.py
SylvainDe: If parentheses are important for these tests (which is likely), maybe I could replace the

    print expression

with

    raise expression

or

    assert expression

PR #670, opened 2017-07-31
Be precise about the noqa comment

mitar:
pydocstyle uses them as well, but they allow customization like `# noqa: D102`. We do not want those comments to enable unconditional noqa in pycodestyle.

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -1420,7 +1420,7 @@ else:
         """Read the value from stdin."""
         return TextIOWrapper(sys.stdin.buffer, errors='ignore').read()
 
-noqa = lru_cache(512)(re.compile(r'# no(?:qa|pep8)\b', re.I).search)
+noqa = lru_cache(512)(re.compile(r'# no(?:qa|pep8)(?!:)\b', re.I).search)
 
 
 def expand_indent(line):


pycodestyle.py
```diff
@@ -1420,7 +1420,7 @@ def stdin_get_value():
         """Read the value from stdin."""
         return TextIOWrapper(sys.stdin.buffer, errors='ignore').read()
 
-noqa = lru_cache(512)(re.compile(r'# no(?:qa|pep8)\b', re.I).search)
+noqa = lru_cache(512)(re.compile(r'# no(?:qa|pep8)(?!:)\b', re.I).search)
```
hoylemd: This breaks backwards compatibility.

mitar: In which way? Do you think that somebody who is doing `noga: custom-check` expects to disable all checks?

If you have `# noqa`, without `:`, it works as it had worked until now.

mitar:
> Existing functionality supports bare noqa, and while I can't speak for the entire community, I know many codebases that use bare noqa.

Yes, and this code continues to work for them. But some other programs (like pydocstyle) uses non-bare noqa.

Maybe I am misunderstanding my regex. :-) The idea with negative lookahead was that if there is a `:`, it should not match it, but if there is no `:` it should still match it. So if you use only `# noqa` it works as it is working currently, but if there is `# noqa: directive` it does not turn off all checks.

pycodestyle.py
hoylemd: Huh. you're right.  I'm not sure why though.  There's no `?` or `*` after the group, so I'd expect it to see >0 `:` at the end.  Sorry, I guess I haven't come across the `?! prefix` before.

sigmavirus24:
So prior to Flake8 3.0 which added support for `# noqa: violation1,violation2,...` people were already using that syntax to *document* which violations they were ignoring in-line. With that context, this does in fact break backwards compatibility. Now, people who only use `pycodestyle` but still want to document what/why they're ignoring will be broken. I agree that the situation you've encountered is sub-optimal, but I'd suggest you use slightly more advanced tooling (i.e., Flake8) which will incorporate pycodestyle's checks and can understand the situation and behave appropriately.

asottile:
closing given the above -- thanks for the patch nonetheless!

PR #672, opened 2017-08-03
Propose .pycodestyle and .pycodestylerc as project level config files

taras80:


sigmavirus24:
Hi @taras80!

Thanks for sending a pull request. Unfortunately, this project *explicitly* removed support for `.<projectname>` config files some time ago because they're messy, confusing, and interact poorly with the other files that we allow configuration to be stored in. Even if we did want to add that support *back* I don't think we'd simultaneously add that support back **and** add support for *N* more such filenames.

Please feel free to look at unassigned issues if you'd like to make other contributions to this project.

Cheers,
Ian

Issue #344, opened 2014-11-07
Enforce 72 character docstring line width


mblayman:
PEP8 states:

> For flowing long blocks of text with fewer structural restrictions (docstrings or comments), the line length should be limited to 72 characters.

The current implementation only enforces the maximum line length (`E501`). I'd like to suggest the addition of an `E503` error to enforce the restriction of these "flowing long blocks of text."

timj:
I would really welcome this test as we have continuing arguments on our project about docstring lengths (I'd like comment blocks to be restricted to that same length as well) vs other line lengths and distinguishing the two (as is done in PEP8) would be incredibly helpful. I understand backwards compatibility issues but can I suggest that we deal with this by having a new config parameter for `max-docstring-length` and default it to whatever `max-line-length` is set to. Then existing users would not even notice a change but people would be able to explicitly set `max-docstring-length` to 72 or 79 or whatever and see the E503 errors.

timj:
@sigmavirus24 would you consider accepting a patch as I describe above? This should be a backwards compatible approach.

sigmavirus24:
@timj would it not be better to add this to pydocstyle?

timj:
Hmm. This is something covered by PEP8 so it would seem to be relevant to this. Does pydocstyle get called by flake8? @jonathansick do you have an opinion?

sigmavirus24:
@timj `flake8-docstrings` is what calls pydocstyle from Flake8. Either way, I'd rather not implement this here if it would be more appropriate in pydocstyle or vice-versa.

timj:
pycodestyle explicitly states: "pycodestyle (formerly pep8) is a tool to check your Python code against some of the style conventions in PEP 8."

[PEP 8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) says:

> For flowing long blocks of text with fewer structural restrictions (docstrings or comments), the line length should be limited to 72 characters.

so it doesn't seem unreasonable for the program that checks PEP8 compliance to have checks listed in PEP8.

PEP257 (covered by pydocstyle) doesn't mention line lengths and doesn't mention comment strings. This strongly suggests to me that pycodestyle should be dealing with warnings about docstring and comment line length.

sigmavirus24:
[...]

PR #674, opened 2017-08-15
Support for W505 - doc line length

brianv0:
An implementation of #344. 

This implementation takes the effective default of 79 characters for the default max doc length despite PEP8 recommendations being 72, as well as setting only a warning. In addition, W504 is added to the default ignore list.

Builds tested with py27 and py35 (don't have others on my platform, sorry!)


brianv0:
Just a note: I'm not sure how to best handle the defaults in lines 1607-1609, as well as how to communicate those defaults sensibly to the user.

Instead of saying the default is `MAX_LINE_LENGTH` (= 79), can we say the default is `max-line-length`, so it's clear that, if `max-line-length` is specified, that `max-doc-length` take that default if it's also not specified.

timj:
Does W504 need to be in the default ignore list? If the default line length is always `max-line-length` how can W504 ever appear for people? Why shouldn't it be an enabled warning? Do you get two warnings if a line is longer than max-line-length and max-doc-length?

brianv0:
If you were explicitly ignoring E501 (I imagine there's a large amount of code which does this), then W504 might unexpectedly show up if you upgraded to a new version of pycodestyle and you had long docstrings which were previously ignored.

brianv0:
I redid this so it's absolutely disabled by default, trying to follow behavior changes similar to `--hanging-indent`, but it's enabled in tox and the testsuite.

On a related note, it seems like using `--testsuite` ignores config in setup.cfg.

AWhetter:
Just a heads up, error code W504 conflicts with the same code used in #502

timj:
Thanks. I assume it doesn't matter if W505 turns up even if W504 is missing?

AWhetter:
Yep that would be fine!

timj:
Any thoughts on how we can make progress on this PR?

brianv0:
Okay, so now this branch is conflicting because it hasn't merged. 

I'm happy to fix those conflicts, but it'd be great if we could get some guidance on this.

IanLee1521:
Hi @brianv0 -- I'm getting myself caught up after some time away... Long story short, I am open to this PR, we would need to resolve the conflicts, but also a comment. In your code, you mix 72 and 79 characters as the line length... for docstrings we would want 72 characters max, right? (ref: https://www.python.org/dev/peps/pep-0008/#maximum-line-length)

brianv0:
Yes, the goal is to support 72 character line length docstrings and 79 character line-length by default, but allow developers to expand to 99 characters, which the linked-to section mentions:

>>> For code maintained exclusively or primarily by a team that can reach agreement on this issue, it is okay to increase the nominal line length from 80 to 100 characters (effectively increasing the maximum length to 99 characters

Notably, in that case, you are changing the nominal line length (79-99), but not changing the docstring doc length, which is 72. For my purposes, this pull request is also intended to allow that combination (99 character line length + 72 character docstrings), but it's sufficiently general that users could choose the combinations they care about, and it's turned off by default.

docs/intro.rst
```diff
@@ -402,6 +404,8 @@ This is the current list of error and warning codes:
 +------------+----------------------------------------------------------------------+
 | W503 (*)   | line break occurred before a binary operator                         |
 +------------+----------------------------------------------------------------------+
+| W504 (\*^) | doc line too long (82 > 79 characters)                               |
```
IanLee1521: I'm pretty sure you don't need the `\` .. Edit: though maybe I'm wrong looking up above..

IanLee1521: I'm wondering if this should be an `E503` instead of `W504`, since this isn't really do do with "line breaks".

setup.cfg
```diff
@@ -5,3 +5,4 @@ universal = 1
 select =
 ignore = E226,E24
 max_line_length = 79
+max_doc_length = 79
```
IanLee1521: shouldn't this be 72 ?

tox.ini
```diff
@@ -10,9 +10,9 @@ skip_missing_interpreters=True
 [testenv]
 commands =
 	{envpython} setup.py install
-	{envpython} pycodestyle.py --testsuite testsuite
+	{envpython} pycodestyle.py --max-doc-length=79 --testsuite testsuite
```
IanLee1521: These should probably also be 72, no?

brianv0: If I change this to 72, I need to actually fix a bunch of docstrings in pycodestyle itself (I can do that)

IanLee1521: I think that is the right way to go, that we should eat our own dog food, as it were. :)

docs/intro.rst
brianv0: Should it be an error rather than a warning though? I was trying to be a bit gentler to existing code. We could rename the heading of the W5 section to just say "Line Length" to match the E5 section

setup.cfg
brianv0: Similar to the tox.ini, I seem to recall setting this to 72 will cause it to fail on the current pycodestyle codebase.

timj:
@brianv0 so is the plan to fix the entire package to be PEP8 docstring length compliant?

hoylemd:
If that is the plan, and it just needs to get done, I can volunteer to help :)

brianv0:
I'll get back to this this week.. working a bit on it now.

brianv0:
Hi @IanLee1521 : I added a few more commits (and fixed a few bugs I found 
b8a2b13) In the docstrings, some of the first lines broke into two lines, and I rewrote a few where I thought the meaning was preserved, and a few more I added a `# noqa` where it seemed like the docstrings were necessarily long

timj:
@brianv0 is this ready for review again? @IanLee1521 are you happy?

peterjc:
Is the plan to rename this (documentation line length check) to W405 given W403 and W404 would make a nice pair for the binary operator line breaks (and the apparent consensus that W404 would be used for that)?

timj:
Who gets to decide on the number to use for codes? I have no idea who the release managers for pycodestyle are.

timj:
#502 seems to have been open for more than a year and a half so I'm not sure whether we can rely on it being merged.

brianv0:
I'd be happy to come up with a new number if #502 is merged before this Pull Request, but I think it would need a rebase first in any case, as it doesn't pass checks/conflicts with the base branch. I'm not sure using an out-of-order number makes sense until we know which pull requests are to be merged.

sigmavirus24:
I strongly suspect #502 will be merged prior to this one. I think there is still contention around whether or not this check belongs in pycodestyle at all.

timj:
@brianv0 seems like you should change the code. @sigmavirus24 where can we have the debate over whether this check should be included? I thought we'd already got agreement.

timj:
Any thoughts?

brianv0:
I've rebased/rewritten this PR to refer to a new error code W505 since support for W504 was merged, and reordered this PR to be two separate commits, one for the implementation of W505 and one so that the code base is self-consistent with the W505 implementation itself (limiting docstring length to 72 characters as PEP8 recommends). I also added support to ignore `#!` docstrings on the first line as well.

All other behavior is still as described.

timj:
Thank you for merging this. We are really looking forward to using it and it will make our lives much easier.

PR #676, opened 2017-08-19
Add W605 warning for invalid escape sequences in string literals

### LINKED ISSUE #633: detect invalid escape sequences
It would be nice if pycodestyle could detect and warn on invalid string escape sequences, which trigger deprecation warnings as of py3.6.

jdufresne:
Starting with Python 3.6, invalid escape sequences in string literals are now deprecated. In a future version of Python, invalid escape sequences will be a syntax error. While this deprecation produces a
runtime warning, it only appears if warnings are enabled and the first time the Python source is compiled to byte code. By adding a check to pycodestyle, projects can take advantage of static analysis to catch and fix these future syntax errors.

For more information on the deprecation, see the [Python release notes](https://docs.python.org/3/whatsnew/3.6.html#deprecated-python-behavior)

> A backslash-character pair that is not a valid escape sequence now generates a DeprecationWarning. Although this will eventually become a SyntaxError, that will not be for several Python releases.

Fixes #633


pycodestyle.py
```diff
@@ -1388,6 +1388,59 @@ def python_3000_backticks(logical_line):
...
+        if token_type == tokenize.STRING:
+            quote = text[-3:] if text[-3:] in ('"""', "'''") else text[-1]
+            # Extract string modifiers (e.g. u or r)
+            quote_pos = text.index(quote)
+            prefix = text[:quote_pos].lower()
+            start = quote_pos + len(quote)
+            string = text[start:-len(quote)]
+
+            if 'r' not in prefix:
+                pos = 0
```
sigmavirus24: Why not do:

```py
pos = string.find('\\')
while pos >= 0:
    pos += 1
    if string[pos] not in valid:
        yield (...)
    pos = string.find('\\', pos + 1)
```

jdufresne:
Thanks for the review.

I agree, that solution is simpler. I've made the suggested change.

PR #688, opened 2017-10-09
Add major.minor classifiers

hugovk:
Plus implementation classifiers


hugovk:
Most of these were added in https://github.com/PyCQA/pycodestyle/pull/654, but this at least adds Jython.

sigmavirus24:
We don't actively test on Jython. So for now I'm going to close this.

Issue #572, opened 2016-09-14
Notify users that "pep8" was renamed to "pycodestyle"


Ede123:
There are many places were there is still outdated information refering to "pep8" instead of "pycodestyle" making no mention of the name change (#466, #481) which can be quite misleading.

For example I just installed "pep8" unaware of the fact and wondered why it was outputting a lot of W503s [when it shouldn't](#499). That's when I realized the project had been renamed and I had installed the outdated pep8 version 1.7.0 instead of the current pycodestyle 2.0.0

I'm not familiar with PyPI and don't know if there is a possibility to deprecate packages but it would be a huge help for new users if the name change was announced prominently at [[1]](https://pypi.python.org/pypi/pep8) and [[2]](http://pep8.readthedocs.io/).
An updated package of pep8 with a deprecation warning might be yet another option if that's not too much to ask.

[1] https://pypi.python.org/pypi/pep8
[2] http://pep8.readthedocs.io/

hugovk:
13 months later and I've only just noticed the name change and am still on the old pep8 1.7.0. I've updated (or tried to update) pep8 many times, and a warning would have helped me.

A prominent note has been added to [[2]](https://pep8.readthedocs.io/):

> Note
>
> This package used to be called pep8 but was renamed to pycodestyle to reduce confusion. Further discussion can be found in the issue where Guido requested this change, or in the lightning talk at PyCon 2016 by @IanLee1521: slides video.

Thanks! 

It'd be great to also put one at [[1]](https://pypi.python.org/pypi/pep8) as well.

I've created PR https://github.com/PyCQA/pycodestyle/pull/690 to issue a warning in pep8.

hugovk:
On further investigation, it looks like the warning has already been done in https://github.com/PyCQA/pycodestyle/commit/d4a60037be95b4b5f84a040bc40ce2ca703640bb in in May 2016 in the release/1.7.x branch, but not yet released.

Please could we have a new release of pep8 with that warning? Thank you!

gsnedders:
See also https://github.com/PyCQA/pycodestyle/issues/602 and https://github.com/PyCQA/pycodestyle/issues/603

sigmavirus24:
1.7.1 was released with a runtime warning

hugovk:
Thanks! I think this issue can be closed now.


```
$ pep8 .
/usr/local/lib/python2.7/site-packages/pep8.py:2124: UserWarning:

[...]

Issue #481, opened 2016-02-17
Project rename -> "pycodestyle"


IanLee1521:
Per the discussion in #466, I'll be renaming this repo to "pycodestyle" in the next day or two, I'll close this issue and post to the code-quality mailing list once I do. Consider this my announcement of the pending name change! :)

sigmavirus24:
@IanLee1521 please let me know in advance so I can be around on IRC in case anything goes sideways. (I don't anticipate that, but I'd like to plan for it regardless.)

IanLee1521:
Thanks for the heads up. Guess it's good I got swamped. Why don't I just plan for Monday morning (ish)?

Also, what's a good IRC client for OSX?

The-Compiler:
> Also, what's a good IRC client for OSX?

I've seen people use [Textual](https://www.codeux.com/textual/) much. If you prefer something console-based, probably [weechat](https://weechat.org/).

sigmavirus24:
@IanLee1521 Monday morning-ish sounds good to me. :) https://webchat.freenode.net/ works too as a temporary client while you figure out your tastes. The channel is `##python-code-quality`` and I'm just realizing that should be in the meta documentation ;)

IanLee1521:
Making the change now.

sigmavirus24:
The project has been renamed so I've closed this. The other (related) issue is marked for 2.0 milestone so I'm leaving that open.

PR #690, opened 2017-10-10
Warn that pep8 has been renamed to pycodestyle

hugovk:
Note: PR against 1.7.x branch.

It was suggested in https://github.com/PyCQA/pycodestyle/issues/572 to make an updated package of pep8 with a deprecation warning to make it clear that pep8 has been replaced with pycodestyle.

That issues was opening in September 2016, and they hadn't realised the rename had happened (#466, #481) and were still using the old, outdated tool.

It's now October 2017, and I only just noticed this too!

Here's a suggested warning that could go into an updated pep8 1.7.x to let people know of the change. Perhaps the warning could have wording to make it clear that pep8 is also outdated and no longer maintained.

### DIFF
--- a/pep8.py
+++ b/pep8.py
@@ -54,6 +54,7 @@ import time
 import inspect
 import keyword
 import tokenize
+import warnings
 from optparse import OptionParser
 from fnmatch import fnmatch
 try:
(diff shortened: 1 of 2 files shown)


hugovk:
On further investigation, it looks like this has already been done in https://github.com/PyCQA/pycodestyle/commit/d4a60037be95b4b5f84a040bc40ce2ca703640bb in in May 2016 in the release/1.7.x branch, but not yet released.

Please could we have a new release of pep8 with that warning? Thank you!

sigmavirus24:
1.7.1 has been released with that warning.

PR #694, opened 2017-10-16
Include license file in the generated wheel package

jdufresne:
The wheel package format supports including the license file. This is done using the `[metadata]` section in the `setup.cfg` file. For additional information on this feature, see:

https://wheel.readthedocs.io/en/stable/index.html#including-the-license-in-the-generated-wheel-file

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,5 +1,6 @@
 include *.txt
 include *.rst
+include LICENSE
 recursive-include docs *
 recursive-include testsuite *
 recursive-exclude docs *.pyc
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,9 @@
 [bdist_wheel]
 universal = 1
 
+[metadata]
+license_file = LICENSE
+
 [pycodestyle]
 select =
 ignore = E226,E24


IanLee1521:
Hi @jdufresne, thank you for the suggestion. Instead of putting this in the `setup.cfg` though, how about we add it into the `MANIFEST.in` file, which is where we define all of the other files to be included in the package?

jdufresne:
Good point about `MANIFEST.in`.

I think we should do both as they serve difference purposes. `MANIFEST.in` ensures that the `LICENSE` file is included in the source distribution while the `setup.cfg` change ensures that the LICENSE file is included in the wheel distribution. One or the other isn't sufficient to achieve both. The latest revision includes the `MANIFEST.in` change. I happen to use the wheel distribution and so it was the one I noticed.

IanLee1521:
Ah, I see.

Are there options in the `setup.cfg` file then that would allow for including all of the same files in both package types then? E.g. `CHANGES.txt`, `CONTRIBUTING.rst`, and `README.rst`?

jdufresne:
I don't think so. There is an open issue for wheel to support bundling arbitrary files: https://github.com/pypa/wheel/issues/138

Keep in mind, wheel is a binary packaging format and not intended to be as general purpose as the source distribution. It should normally only contain the files necessary to install the package for running. CHANGES.txt, CONTRIBUTING.rst, and README.rst might be considered outside that scope.

The license file is a bit of an issue as it often contains language similar to:

> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

IanLee1521:
Good to know, thanks for the link. I will go ahead and merge this. Thanks for the contribution!

PR #697, opened 2017-10-24
Fix missed quoting in E722 check

dirkmueller:
The warning looks fairly odd:

E722 do not use bare except'

change this to

E722 do not use bare 'except'


IanLee1521:
Good catch, thank you! 👍

asottile:
👍 noticed this via flake8, was about to suggest the same patch

PR #699, opened 2017-10-25
Fix space in flag name

Wilfred:
Previously, the line break in the source would result in
`--hang- closing` (note the space) being rendered in the HTML.


IanLee1521:
LGTM, thanks @Wilfred !

PR #705, opened 2017-11-06
Add check for over-indented blocks

### LINKED ISSUE #430: No complaint about inconsistent indentation levels
`pep8` neglects to complain about code with mismatched indentation levels, such as this snippet:

```
if True:
        print 'yes'
else:
    print 'no'
```

Example taken from a [complaint](https://stackoverflow.com/questions/32032206/pep8-has-no-option-to-force-same-indentation-in-whole-code) about this problem on StackOverflow.  `pylint` gives a reasonable error here, namely:

> W:  2, 0: Bad indentation. Found 8 spaces, expected 4 (bad-indentation)

The PEP 8 standard says simply:

> Use 4 spaces per indentation level.

And I would like to have a warning about any indentation level other than 4 spaces. Here's the [beginning of a fix](https://github.com/schmiddy/pep8/tree/issue-430). I think these errors might need to be separated based on:

 a. mixing different indentation levels, as in the example above
 b. using any indentation level other than 4 spaces (including tabs) 

IMO a.) should definitely be an error by default. I can see the case for b.) being off by default and/or just a warning, particularly since PEP 8 makes an allowance for tabs by saying:

> Tabs should be used solely to remain consistent with code that is already indented with tabs.

jdufresne:
In a project with all lines indented 4 spaces, I noticed pycodestyle was not catching code that was accidentally indented two levels (8 spaces). The over indentation was unintended and can be caught during static
analysis.

Fixes #430


jdufresne:
Rebased to resolve merge conflicts.

CHANGES.txt
```diff
@@ -20,6 +20,7 @@ New checks:
   in Python 3.7. PR #684.
 * Add E252 error for missing whitespace around equal sign in type annotated
   function arguments with defaults values. PR #717.
+* Add E117 error for over-indented code blocks.
```
sigmavirus24: I don't believe this is the right place to add this because 2.4.0 is already released

testsuite/W19.py
```diff
@@ -109,8 +109,8 @@ def long_function_name(
 '''sometimes, you just need to go nuts in a multiline string
 	and allow all sorts of crap
   like mixed tabs and spaces
-      
-or trailing whitespace  
+
+or trailing whitespace
```
sigmavirus24: I'm pretty sure we wanted to keep those trailing white spaces

jdufresne:
Thanks for the review. I've updated the PR with the suggested changes. :slightly_smiling_face:

sciyoshi:
Hi @jdufresne @sigmavirus24, this change has started causing E117 to be reported for any files containing tabs. My guess is that this change sees them as being 8 spaces and thus being overindented. For now we will simply disable this error, but I think this should be addressed and I'll open a separate issue for this.

PR #706, opened 2017-11-06
Fix handling of diffs with mnemonic prefixes

### LINKED ISSUE #701: Mnemonic prefixes break pycodestyle --diff execution
In the `diff` section in my gitconfig I have an option `mnemonicprefix` set to `true`. This means that git will use `i/w/c/o` instead of `a/b/c/d` in the generated diff. Apparently that breaks execution of `pycodestyle` when `--diff` argument is provided.

Command:

    git diff -U0 --staged | pycodestyle --diff --max-line-length=120

Output:

    ./i/path.py:1:1: E902 IOError: [Errno 2] No such file or directory: './i/path.py'

To fix it I could use `--no-prefix` in git but I am wondering whether this should be handled by `pycodestyle` itself.

mic4ael:
fixes #701

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -1538,7 +1538,9 @@ def parse_udiff(diff, patterns=None, parent='.'):
             rv[path].update(range(row, row + nrows))
         elif line[:3] == '+++':
             path = line[4:].split('\t', 1)[0]
-            if path[:2] == 'b/':
+            # Git diff will use (i)ndex, (w)ork tree, (c)ommit and (o)bject
+            # instead of a/b/c/d as prefixes for patches
+            if path[:2] in ('b/', 'w/', 'i/'):
                 path = path[2:]
             rv[path] = set()
     return dict([(os.path.join(parent, path), rows)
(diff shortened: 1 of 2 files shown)


mic4ael:
Could anyone please take a look at this?

pycodestyle.py
```diff
@@ -1538,7 +1538,7 @@ def parse_udiff(diff, patterns=None, parent='.'):
             rv[path].update(range(row, row + nrows))
         elif line[:3] == '+++':
             path = line[4:].split('\t', 1)[0]
-            if path[:2] == 'b/':
+            if path[:2] in ('b/', 'w/', 'i/'):
```
IanLee1521: @mic4ael -- would you be able to add a comment here to explain where each of these values are coming from? That way we know why those are the magic values we want. As an aside, should this be "any character followed by a /" ? Something like: `if path[0].isalpha() and path[1] == '/':`

Additionally, if I could request that you add some tests for these cases into the test suite, then I think this should be mergeable (I don't see any issues, @sigmavirus24 , could you please be a second opinion?)

mic4ael: I am not sure whether the approach of "any character followed by a /" is the right one since in `git` the list of used prefixes is limited (as far as i know). Sure, I will add the tests ;)

mic4ael:
Updated ;)

mic4ael:
Rebased to latest master

PR #527, opened 2016-06-02
[DRAFT] Add support for checks on unconsistent returns

ArcTanSusan:
The original author is initially SylvainDe in https://github.com/PyCQA/pycodestyle/pull/431, which is a PR that has been inactive for over a year. In PyCon 2016 sprints, I've rebased & fixed all merge conflicts. The unit tests all pass.

PR #711, opened 2017-12-02
Check inconsistent return statements

r3m0t:
This implements the same as #431  #527 (requested in #399). This is about the best that can be done without importing the `ast` module, and therefore this PR is much smaller.

PEP8:

> Be consistent in return statements. Either all return statements in a
function should return an expression, or none of them should. If any return
statement returns an expression, any return statements where no value is
returned should explicitly state this as return None, and an explicit
return statement should be present at the end of the function (if
reachable).

This:

> Either all return statements in a
function should return an expression, or none of them should. If any return
statement returns an expression, any return statements where no value is
returned should explicitly state this as return None.


r3m0t:
Things I'm not sure about:

* Should it be disabled by default? It looks like it'll trip up on a lot of codebases (including, ironically, this one)
* Should it support `noqa`?

pycodestyle.py
```diff
@@ -495,6 +495,58 @@ def indentation(logical_line, previous_logical, indent_char,
...
+    as return None, [and an explicit return statement should be present at the
+    end of the function (if reachable).]
+
+    The reachability constraint is not implemented due to its complexity.
+
+    Okay: def a():\n    return 1
+    Okay: def a():\n    return 1\n    return 2
+    Okay: def a():\n    return
+    Okay: def a():\n    return\n    return
+    Okay: def a():\n    def b():\n        return\n    return b
```
r3m0t: These can be taken out as they are tested in the more detailed tests.

sigmavirus24:
So sorry for the delay on this @r3m0t:

> Should it be disabled by default? It looks like it'll trip up on a lot of codebases (including, ironically, this one)

That makes me lean towards yes.

> Should it support noqa?

I think not. Mostly because it will be disabled by default, it shouldn't be required here, plus once it's on and a code-base is passing, I somehow doubt folks will want to allow someone to `# noqa` it. We can always revise and add in later.

> Should it emit no more than one warning for a function?

Eh. I don't think we have other checks in this codebase that behave that way so it's fine that this emits more than one per function.

bz2:
@sigmavirus24 Can this just be landed then? Was looking for the pycodestyle code for this rule, and found it was added for pylint some time ago https://github.com/PyCQA/pylint/issues/1267 but never done here.

PR #720, opened 2018-01-17
Remove support for EOL Python 2.6 and 3.3

jdufresne:
Python 2.6 an 3.3 are end of life. They are no longer receiving bug fixes, including for security issues. Python 2.6 went EOL on 2013-10-29 and 3.3 on 2017-09-29. For additional details on support Python versions, see: https://devguide.python.org/#status-of-python-branches
    
Removing support for EOL Pythons will reduce testing and maintenance resources. Removed all workarounds for older Pythons.
    
Updated trove classifiers and documentation to better communicate supported Python versions.
    
Additionally, pass python_requires argument to setuptools. Helps pip decide what version of the library to install. 
    
https://packaging.python.org/tutorials/distributing-packages/#python-requires
    
> If your project only runs on certain Python versions, setting the python_requires argument to the appropriate PEP 440 version specifier string will prevent pip from installing the project on other Python versions.
    
https://setuptools.readthedocs.io/en/latest/setuptools.html#new-and-changed-setup-keywords
    
> python_requires
>
> A string corresponding to a version specifier (as defined in PEP 440) for the Python version, used to specify the Requires-Python defined in PEP 345.
    
Can now use more modern Python syntax including dictionary comprehension as well as more generators.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,6 +34,7 @@ setup(
     namespace_packages=[],
     include_package_data=True,
     zip_safe=False,
+    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*',
     install_requires=[
         # Broken with Python 3: https://github.com/pypa/pip/issues/650
         # 'setuptools',
@@ -51,10 +52,8 @@ setup(
         'Operating System :: OS Independent',
         'Programming Language :: Python',
         'Programming Language :: Python :: 2',
-        'Programming Language :: Python :: 2.6',
         'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.3',
         'Programming Language :: Python :: 3.4',
         'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,12 +6,8 @@ install:
 script: tox
 matrix:
   include:
-    - python: 2.6
-      env: TOXENV=py26
     - python: 2.7
       env: TOXENV=py27
-    - python: 3.3
-      env: TOXENV=py33
     - python: 3.4
       env: TOXENV=py34
     - python: 3.5
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+UNRELEASED
+----------
+
+Changes:
+
+* Remove support for EOL Python 2.6 and 3.3.
+
 2.4.0 (2018-04-10)
 ------------------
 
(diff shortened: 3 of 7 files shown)


setup.py
```diff
@@ -34,6 +34,7 @@ def get_long_description():
     namespace_packages=[],
     include_package_data=True,
     zip_safe=False,
+    python_requires='>=2.6, !=3.0.*, !=3.1.*, !=3.2.*',
```
sigmavirus24: I would rather this be `>= 2.7` also `!= 3.3.*`

IanLee1521: I am inclined to leave this as proposed (`python_requires='>=2.6, !=3.0.*, !=3.1.*, !=3.2.*'`), as that matches (in addition to our trove classifiers) what we test for with tox and travis.

@sigmavirus24, what is your rationale for removing 2.6 and 3.3.x ? I am inclined to want to leave them as long as they don't add any additional maintenance overhead or prevent new development...

sigmavirus24: 2.6 and 3.3 are no longer supported by the packaging tools we rely on. setuptools, wheel, and other libraries (I think even pip) have dropped support for these EOL'd versions of Python. We will need additional work to actually support them in our CI and other systems. I don't think it's worth the work any longer.

Further, using BigQuery we can find the most popular downloads:

```
❯❯❯ pypinfo -d 90 -pc -md pycodestyle pyversion
```

| python_version | percent | download_count |
| -------------- | ------: | -------------: |
| 2.7            |  44.77% |      1,015,324 |
| 3.6            |  34.77% |        788,557 |
| 3.5            |  12.99% |        294,623 |
| 3.4            |   6.06% |        137,371 |
| 3.7            |   0.45% |         10,286 |
| 3.3            |   0.36% |          8,128 |
| 2.6            |   0.28% |          6,277 |
| None           |   0.24% |          5,509 |
| 1.17           |   0.06% |          1,449 |
| 3.2            |   0.03% |            580 |

3.3 and 2.6 make up 0.66% of all downloads over the last 90 days

jdufresne: @sigmavirus24 Can you clarify if you'd like me to address removal of 2.6 & 3.3 universally or just this one line? I think if the projects official stance is 2.6 & 3.3 are not supported, workarounds and documentation should be removed and updated completely to make this clear to the user. I have no personal needs to continue supporting 2.6 & 3.3, so I don't mind doing this additional work. Please let me know.

IanLee1521: @sigmavirus24 -- good points. Thank you for those.

@jdufresne  -- If you can update this PR to remove all the Python 2.6 and 3.3 references / support, I think that is the direction we want to go. (tox.ini, setup.py, docs/, etc)

jdufresne:
I have updated the PR, commit message, code, and documentation to remove all 2.6 & 3.3 support. Thanks!

PR #721, opened 2018-01-17
Allow indenting chained method calls inside parentheses

jdufresne:
For example, a Django queryset can have many chained method calls (filter, exclude, count, etc.) as well as long arguments. To avoid long lines or ending lines with a backslash, it may be wrapped in parentheses.

```py
user_count = (
    User.objects
        .filter(really_long_filter='thing1')
        .exclude(really_long_exclude='thing2')
        .count())
```

Or with a hanging parenthesis:

```py
user_count = (
    User.objects
        .filter(really_long_filter='thing1')
        .exclude(really_long_exclude='thing2')
        .count()
)
```

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -619,7 +619,11 @@ def continued_indentation(logical_line, tokens, indent_level, hang_closing,
                 elif indent[depth]:
                     error = "E127", "over-indented for visual indent"
                 elif not close_bracket and hangs[depth]:
-                    error = "E131", "unaligned for hanging indent"
+                    if token_type == tokenize.OP and text == '.':
+                        # Allow indenting chained method calls.
+                        continue
+                    else:
+                        error = "E131", "unaligned for hanging indent"
                 else:
                     hangs[depth] = hang
                     if hang > 4:
--- a/testsuite/E12.py
+++ b/testsuite/E12.py
@@ -373,4 +373,13 @@ print(True)
 
 print(a
 , end=' ')
-#:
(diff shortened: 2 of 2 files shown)


sigmavirus24:
While we're not named `pep8` we still heavily work on implementing PEP-0008 the document. As I recall, this has no basis in the document for this change. This could be a plugin to Flake8 though.

hoylemd:
@sigmavirus24 Does flake8 still wrap pycodestyle entirely?  I've been using both, but if I can just use one I'd probably be a much happier coder :p


(Because then I only need to maintain one tool/integration with my environment)

hoylemd:
@sigmavirus24 Oh, sure.

I'm wondering if flake8 calls pycodestyle as part of it's analysis, or otherwise covers all of the same types of issues.

Put another way: Do any errors exist such that: `pycodestyle` detects them AND `flake8` does not detect them?

sigmavirus24:
> I'm wondering if flake8 calls pycodestyle as part of it's analysis

Flake8 calls the pycodestyle checks directly as if they're plugins. It's core engine is an improved version of the one in pycodestyle.

> Do any errors exist such that: pycodestyle detects them AND flake8 does not detect them?

That I'm aware of? No. Is it plausible? Certainly

hoylemd:
@sigmavirus24 Great answers, thanks!  I think I can safely use flake8 on it's own then.  I didn't expect that the checks would be called directly, but of course that's a better way to do it. :)

PR #722, opened 2018-01-19
Also accept string lists as "E201,E202" from **kwargs arguments

evandrocoan:
The package python-language-server was passing the *args as string
while internally it expects an list, leading to the line:
```python
            # Ignore all checks which are not explicitly selected
            options.ignore = ('',) if options.select else tuple(options.ignore)
```
make the arguments being interpreted as:
```
options.ignore: ('E', '2', '0', '1', ',', 'E', '5', '0', '1')
```
instead of just `options.ignore: ('E201', 'E501')`

Issue:
1. https://github.com/tomv564/LSP/issues/244 pycodestyle plugin ignore setting ignored:

For this was required to set some default values to ensure the type assertions.

And also added a new example for ignore and select options, showing
you can also pass rules as E201.

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -1538,7 +1538,9 @@ def parse_udiff(diff, patterns=None, parent='.'):
             rv[path].update(range(row, row + nrows))
         elif line[:3] == '+++':
             path = line[4:].split('\t', 1)[0]
-            if path[:2] == 'b/':
+            # Git diff will use (i)ndex, (w)ork tree, (c)ommit and (o)bject
+            # instead of a/b/c/d as prefixes for patches
+            if path[:2] in ('b/', 'w/', 'i/'):
                 path = path[2:]
             rv[path] = set()
     return dict([(os.path.join(parent, path), rows)
--- a/testsuite/test_shell.py
+++ b/testsuite/test_shell.py
@@ -191,3 +191,13 @@ class ShellTestCase(unittest.TestCase):
         self.assertFalse(errcode)
         self.assertFalse(stdout)
         self.assertFalse(stderr)
+
+        for index, diff_line in enumerate(diff_lines, 0):
+            diff_line = diff_line.replace('a/', 'i/')
+            diff_lines[index] = diff_line.replace('b/', 'w/')
+
+        self.stdin = '\n'.join(diff_lines)
+        stdout, stderr, errcode = self.pycodestyle('--diff')
+        self.assertFalse(errcode)
+        self.assertFalse(stdout)
+        self.assertFalse(stderr)


pycodestyle.py
```diff
@@ -2056,6 +2056,25 @@ def __init__(self, *args, **kwargs):
             if 'paths' in options_dict:
                 self.paths = options_dict['paths']
 
+        # After options.__dict__.update we need to ensure the *args and **kwargs where correctly passed
+        assert isinstance(options.verbose, int)
```
sigmavirus24: This disappear with `python -O` so I don't think this does what you want.

evandrocoan: I just ran `python3 -O pycodestyle.py --testsuite testsuite` and `verbose` option was present on the debugging statements I had added:
```Python
StyleGuide: args: ()
StyleGuide: kwargs: {'parse_argv': True}
StyleGuide: process_options, select: []
StyleGuide: process_options, ignore: []
StyleGuide: process_options, select: []
StyleGuide: process_options, ignore: []
StyleGuide: options: {'show_source': False, 'statistics': False, 'select': [], 'quiet': 0, 'filename': ['*.py'], 'count': False, 'benchmark': False, 'repeat': True, 'testsuite': 'testsuite', 'reporter': None, 'exclude': ['.svn', 'CVS', '.bzr', '.hg', '.git', '__pycache__', '.tox'], 'verbose': 0, 'format': 'default', 'ignore': [], 'show_pep8': False, 'diff': False, 'max_line_length': 79, 'hang_closing': False, 'config': False, 'doctest': False}
StyleGuide: options.ignore 1: []
StyleGuide: options.ignore 2: ()
3061 lines tested: 37 files, 381 test cases.
Test passed.
```

But I already had created this function:
```python
File: pycodestyle/pycodestyle.py
2461: def customisinstance(values_object, option_name, target_type):
2462:     """ Used to check optional options like verbose which is not present
2463:     when running with `python -o`"""
2464:     values_dict = values_object.__dict__
2465:     if option_name in values_dict:
2466:         return isinstance(values_dict[option_name], target_type)
2467:     return True
```

pycodestyle.py
```diff
@@ -2411,6 +2388,38 @@ def _parse_multi_options(option, split_token=','):
...
+                     ("show_source", bool),
+                     ("show_pep8", bool),
+                     ("statistics", bool),
+                     ("count", bool),
+                     ("hang_closing", bool),
+                     ("diff", bool),
+                     ("benchmark", bool),)
+    for option, default_type in default_types:
+        if not isinstance(options_dict[option], default_type):
+            raise ValueError("The option `%s=%s` must be of the type `%s` "
```
sigmavirus24: This is mildly better but I really don't know that we should be doing strict type-checking here.

evandrocoan: Then it was doing what I expected. I do not like the ideia of type checking while running on production environment to not eat CPU cycles unnecessarily.

The assert statements are meant to the development environment, and for the ones this this directly by its API, as `python-language-server`

sigmavirus24:
So I read through the original issue. This appears to be (by that maintainer's perspective) a bug in Pyls, not in Pycodestyle. In otherwords, Pyls is passing the wrong values and it seems to be the only consumer of pycodestyle doing things this way. Please file an issue there and we can follow along.

evandrocoan:
I already pushed a pull request fixing the issue there:

1. https://github.com/palantir/python-language-server/pull/234 The error report must explicitly call _ignore_code() before accept 

The issue here is that they pass the value as `E201,E202` and here you do 

https://github.com/PyCQA/pycodestyle/blob/534310014a1dad2a5aeca486fa2dccda93eb930b/pycodestyle.py#L2107-L2108

Which puts the string `E201,E202` make make this line:

https://github.com/PyCQA/pycodestyle/blob/534310014a1dad2a5aeca486fa2dccda93eb930b/pycodestyle.py#L2125

To turn it into `('E', '2', '0', '1', ',', 'E', '5', '0', '1')`.

___

There are 2 options:

1. Force everybody to pass `"E201,E202"` by command line and a list by API `["E201","E202"]`
1. Or let everybody to pass `"E201,E202"`  or a list ["E201","E202"] by command line or by API.

On this pull request I implement the option 2. You can either pass a list or a string and you code will work.

What you are suggesting is to close this pull request and force them to pass a list `["E201","E202"]` instead of string `"E201,E202"`.

Would be much better if everybody could be pass only a string `"E201,E202"`, either by the command line, or by the API internally by `**kwargs`:

https://github.com/PyCQA/pycodestyle/blob/534310014a1dad2a5aeca486fa2dccda93eb930b/pycodestyle.py#L2095-L2102

___

I do not know why by you seem to do not like the fact of using `assert` juding by your comment here:

> This is mildly better but I really don't know that we should be doing strict type-checking here.

Then I removed the asserts for good. But they are doing no harm and only would help future developer of this project.

Why do you not agree in allowing the list options to be either `"E201,E202"` or `["E201","E202"]` when updating the dict on line 2108?

https://github.com/PyCQA/pycodestyle/blob/147c399c357a08fa0af982161024af66710e35e5/pycodestyle.py#L2108

Currently, it is only accepting lists. Everybody else who passed the list arguments as string `"E201,E202"` instead of python list ["E201","E202"] will have their code turned into `('E', '2', '0', '1', ',', 'E', '5', '0', '1')`.

This fixes this by also doing the list parsing for the `**kwargs` values after the dict update:
https://github.com/PyCQA/pycodestyle/blob/147c399c357a08fa0af982161024af66710e35e5/pycodestyle.py#L2108

evandrocoan:
The change is call this function:
```
def _parse_list_multi_options(options):
    options.filename = _parse_multi_options(options.filename)
    options.exclude = normalize_paths(options.exclude)
    options.select = _parse_multi_options(options.select)
    options.ignore = _parse_multi_options(options.ignore)
```
After doing this:
https://github.com/PyCQA/pycodestyle/blob/147c399c357a08fa0af982161024af66710e35e5/pycodestyle.py#L2108

Then it will parse the list options `"E201,E202"` as list `["E201","E202"]`, either they come from the command line CLI, or by the `**kwargs` parameter.

Issue #728, opened 2018-02-11
FutureWarning: Possible nested set for EXTRANEOUS_WHITESPACE_REGEX


The-Compiler:
When running pycodestyle 2.3.1 with Python 3.7, I get:

```
/home/florian/proj/qutebrowser/git/.tox/flake8/lib/python3.7/site-packages/pycodestyle.py:113: FutureWarning: Possible nested set at position 1
  EXTRANEOUS_WHITESPACE_REGEX = re.compile(r'[[({] | []}),;:]')
```

adiroiban:
See https://docs.python.org/dev/whatsnew/3.7.html 

Support of nested sets and set operations in regular expressions as ..... To avoid a warning escape them with a backslash. 

Let me see if I can fix this :)

RawrBear:
Getting the same thing in Atom while using Python 3.7.
Installed linter-flake8 and got 
``python37\lib\site-packages\pycodestyle.py:113: FutureWarning: Possible nested set at position 1
  EXTRANEOUS_WHITESPACE_REGEX = re.compile(r'[[({] | []}),;:]')``
As a python noob, I really don't know what else to add :(

Arcanemagus:
@RawrBear Make sure that your `pycodestyle` version is v2.4.0 or higher as that was the first release that included the fix.

RawrBear:
@Arcanemagus I didn't get notified about your message until now. 
Thanks for the fix. I updated `pycodestyle` through pip which worked but broke `flake8` because apparently its only compatible with version 2.3.1. so linter-flake8 is a no go right now :/

Fix one problem, find another ten lol

jeidsath:
@RawrBear I had the same problem, so I just copied the escaped version of the line to my code:

Fixed version:  https://github.com/PyCQA/pycodestyle/blob/master/pycodestyle.py#L133

declantraynor:
@RawrBear I've found in the meantime that installing `flake8` from master works around the problem:

`pip install -e git+https://gitlab.com/pycqa/flake8#egg=flake8`

Hope this helps.

bartek-rychlicki-regalix:
@declantraynor Maybe I'm nitpicking but `pip install -e` without a specific commit is prone to changes and breakage. I think it's a good idea to pin a specific commit, the newest is `9631dac5` as of time of writing

```
pip install -e git+https://gitlab.com/pycqa/flake8@9631dac5#egg=flake8
```

[...]

PR #733, opened 2018-03-19
Add variables for blank lines.

### LINKED ISSUE #732: Add configuration for blank lines.
I know that a lot of Python code is using a single blank line for methods and 2 blank lines for classes.

Still, I would like to see a flag, to switch to 2 blank lines for methods and 3 blank lines for classes.
In this way, you can use 1 blank line inside a method and don't confuse it with a method delimiter.

This style is used by the Twisted project and for a couple of years Twisted was using a fork which changes the blank_lines function in pycodestyle.

I am starting this discussion to see whether pycodestyle maintainer are happy/willing to help with having support for this upstream, so that we don't have to keep the fork :)

Here is a patch, which extracts the blank lines values into local variables (for now) https://github.com/PyCQA/pycodestyle/compare/master...adiroiban:configurable-blank-lines

If such a functionality is accepted, I am happy to write the implementation and tests.

Thanks a lot for your work. Pycodestyle is of great help in keeping Twisted code clean :)

adiroiban:
Scope
====

This is the initial PR for trying to fix #732 

This does not add public configuration, but just extracts the blank lines configuration into variables.

I hope that these changes are helpful and will make the code easier to read.

I am pushing this simple PR to trigger some feedback.

To solve the blank-lines issue, I could ignore all E3XX errors and them register my own blank_lines function which would be a fork from the pycodestyle version.
But I prefer to reduce the level of forking :)

Changes
=======

For a start, it just extracts the variables into a top level dict which can be easily overwritten.

How to test
=========

Tox tests seem ok,  but I don't know how to update the other tests.

If I run pycodedstle on E30not.py it gave me errors in master:

```
$ git top
eb91b79 Ian Stapleton Cordasco 2018-02-17 Merge pull request #684 from jdufresne/async-await
$ pycodestyle testsuite/E30not.py 
testsuite/E30not.py:6:1: E302 expected 2 blank lines, found 1
testsuite/E30not.py:10:1: E302 expected 2 blank lines, found 0
testsuite/E30not.py:14:1: E302 expected 2 blank lines, found 0
testsuite/E30not.py:17:1: E302 expected 2 blank lines, found 0
testsuite/E30not.py:136:1: E302 expected 2 blank lines, found 0
testsuite/E30not.py:162:1: E302 expected 2 blank lines, found 0
(build) 
```

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -328,12 +328,18 @@ def blank_lines(logical_line, blank_lines, indent_level, line_number,
     E305: def a():\n    pass\na()
     E306: def a():\n    def b():\n        pass\n    def c():\n        pass
     """
+    top_level_lines = 3
+    method_lines = 2
+
     if line_number < 3 and not previous_logical:
         return  # Don't expect blank lines before the first line
     if previous_logical.startswith('@'):
         if blank_lines:
             yield 0, "E304 blank lines found after function decorator"
-    elif blank_lines > 2 or (indent_level and blank_lines == 2):
+    elif (
+        blank_lines > method_lines + 1 or
+        (indent_level and blank_lines == method_lines + 1)
+            ):
         yield 0, "E303 too many blank lines (%d)" % blank_lines
     elif STARTSWITH_TOP_LEVEL_REGEX.match(logical_line):
         if indent_level:
@@ -353,12 +359,18 @@ def blank_lines(logical_line, blank_lines, indent_level, line_number,
                         "nested definition, found 0"
                 else:
                     yield 0, "E301 expected 1 blank line, found 0"
-        elif blank_before != 2:
-            yield 0, "E302 expected 2 blank lines, found %d" % blank_before
-    elif (logical_line and not indent_level and blank_before != 2 and
-          previous_unindented_logical_line.startswith(('def ', 'class '))):
-        yield 0, "E305 expected 2 blank lines after " \
-            "class or function definition, found %d" % blank_before
+        elif blank_before != top_level_lines:
+            yield 0, "E302 expected %s blank lines, found %d" % (
+                top_level_lines, blank_before)
+    elif (
+        logical_line and
+        not indent_level and
+        blank_before != top_level_lines and
+        previous_unindented_logical_line.startswith(('def ', 'class '))
+            ):
+        yield 0, "E305 expected %s blank lines after " \
+            "class or function definition, found %d" % (
+                top_level_lines, blank_before)
 
 
 @register_check
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -95,6 +95,13 @@ except ImportError:
 PROJECT_CONFIG = ('setup.cfg', 'tox.ini')
 TESTSUITE_PATH = os.path.join(os.path.dirname(__file__), 'testsuite')
 MAX_LINE_LENGTH = 79
+# Number of blank lines between various code parts.
+BLANK_LINES_CONFIG = {
+    # Top level class and function.
+    'top_level': 2,
+    # Methods and nested class and function.
+    'method': 1,
+}
 REPORT_FORMAT = {
     'default': '%(path)s:%(row)d:%(col)d: %(code)s %(text)s',
     'pylint': '%(path)s:%(row)d: [%(code)s] %(text)s',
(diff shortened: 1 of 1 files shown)


pycodestyle.py
```diff
@@ -328,12 +335,17 @@ def blank_lines(logical_line, blank_lines, indent_level, line_number,
     E305: def a():\n    pass\na()
     E306: def a():\n    def b():\n        pass\n    def c():\n        pass
     """
+    top_level_lines = BLANK_LINES_CONFIG['top_level']
+    method_lines = BLANK_LINES_CONFIG['method']
```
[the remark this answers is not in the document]
adiroiban: Thanks for the review. Will update it.

I will look at updating all the code. I was not sure this will be accepted and it was a quick hack to demonstrate the direction.

--------

I was expecting that for each instance where pycodestyle asks for 2 blank line, for Twisted we will have 3 :)..and for all other insances where pycodestyle asks for 1 blank line for Twisted there will be 2... reusing the same code number.

This is how it is used now... but with a hardcore fork which is a bit hard to maintain :) https://github.com/twisted/twistedchecker/blob/master/twistedchecker/checkers/pycodestyleformat.py#L218

----------

I was thinking to just disable E3XX from pycodestyle and register a new checker with custom code but this would lead to a lot of code duplication.

This is why I am proposing this path :)

adiroiban:
btw... I saw that since 3 builds ago, travis-ci is failing with python 3.7 https://travis-ci.org/PyCQA/pycodestyle/branches

pycodestyle.py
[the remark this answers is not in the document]
sigmavirus24: I totally understand the _why_ behind this, and agree with it. I wonder if it would be plausible to have some tests which modify these values and assert compliant code, outside of the testsuite.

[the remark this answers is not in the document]
adiroiban: Yes. I want to write tests... no doubt about it... is just that I am confused where and how I am expected to write the tests... see testsuite/E30not.py  which fails on master.

I was thinking to create a new testsuite/test_blank_lines.py file where to write some integration tests for the blank lines code.

I am open to any other suggestion.

adiroiban:
Hi @sigmavirus24 

Please check the initial tests and let me know if this is the right direction.
I am happy to continue writing more tests for the blank_lines code.

I added a diff-cover tox environment to help with checking that changes are covered by the test.
Let me know if you would like to see codecov.io reports. I can help set that up.

I have executed flake8 and found a few issues... one was minor, regarding variable names, and I fixed that.

Travis-CI fails, due to Python2.7 changes.

```
pycodestyle.py:131: FutureWarning: Possible nested set at position 1
  EXTRANEOUS_WHITESPACE_REGEX = re.compile(r'[[({] | []}),;:]')
```

adiroiban:
The python2.7 was reported in #728

tox.ini
```diff
@@ -4,7 +4,7 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py26, py27, py32, py33, py34, py35, py36, pypy, pypy3, jython
+envlist = py26, py27, py32, py33, py34, py35, py36, pypy, pypy3, jython, patchdiff
```
sigmavirus24: Why are you adding patchdiff here? Please, let's keep this PR focused.

pycodestyle.py
```diff
@@ -1665,8 +1682,8 @@ def parse_udiff(diff, patterns=None, parent='.'):
             if path[:2] in ('b/', 'w/', 'i/'):
                 path = path[2:]
             rv[path] = set()
-    return dict([(os.path.join(parent, path), rows)
-                 for (path, rows) in rv.items()
+    return dict([(os.path.join(parent, child), rows)
```
sigmavirus24: Why are you changing this here?

testsuite/support.py
```diff
@@ -83,6 +83,25 @@ def print_results(self):
         print("Test failed." if self.total_errors else "Test passed.")
 
 
+class InMemoryReport(BaseReport):
```
sigmavirus24: :+1:

pycodestyle.py
adiroiban: I run flake8 and found this error... so  I went for this drive by... is 1 line :)

I have reverted this.

testsuite/support.py
```diff
@@ -83,6 +83,25 @@ def print_results(self):
...
+    """
+    Collect the results in memory, without printing anything.
+    """
+
+    def __init__(self, options):
+        BaseReport.__init__(self, options)
```
hoylemd: Why not use `super`? python 2/3 compatability?

sigmavirus24: If BaseReport is not a new-style class I think you need to do this. I don't recall what kind of class it is, though

adiroiban: I will do super :)  BaseReport is new style (class BaseReport(object):)

I just remembered that pycodestuly runs with a log of python versions...and while working on this, I was lazy and just went with this syntax. Will change to super.

Thanks:)

sigmavirus24: I think 2.6+ supports new-style classes + super so you're safe :)

adiroiban: Will use super :)

@sigmavirus24 , can you please confirm whether the new tests are the type of tests that you would like to see for this change?

If you are happy with this style of testing, I can continue to write more test so that they will touch all braches.

Thanks!

sigmavirus24: Yeah, these tests look fantastic, @adiroiban :)

adiroiban:
I didn't had much time...  and I kind of lost interest into manual codestyles and start considering using yapf :)

The unit tests have 100% coverage for the code and I tried to import the main examples from E30.py

As a drive by I have updated the documentation about tests styles.
As a first time contributor, I was a bit confused about how to write and run the tests.

I don't like the custom framework as it make it hard to debug the code.
With unittest style, I can pinpoint a single test and inspect the isolated execution.

I think that this can be ready to merge.

PR #735, opened 2018-03-24
Fix support for Python 3.7.

adiroiban:
Scope
=====

This started as a fix for the #728 issue, where we got a warning and tests are failing on python 3.7.

But I have retargeted for the whole python 3.7

Why we got this
============

In python 3.7 we got `async` keyword.

Changes
=======

Just explicitly escape the nested brackets. I went for escaping both the start and the end.

How to test
==========

Give it a try on Python 3.7... but I hope that Travis-CI will help with the tests.

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -413,7 +413,7 @@ def whitespace_around_keywords(logical_line):
             yield match.start(1), "E272 multiple spaces before keyword"
 
         if '\t' in after:
-            yield match.start(2), "E273  tab after keyword"
+            yield match.start(2), "E273 tab after keyword"
         elif len(after) > 1:
             yield match.start(2), "E271 multiple spaces after keyword"
 
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -121,7 +121,7 @@ RAISE_COMMA_REGEX = re.compile(r'raise\s+\w+\s*,')
 RERAISE_COMMA_REGEX = re.compile(r'raise\s+\w+\s*,.*,\s*\w+\s*$')
 ERRORCODE_REGEX = re.compile(r'\b[A-Z]\d{3}\b')
 DOCSTRING_REGEX = re.compile(r'u?r?["\']')
-EXTRANEOUS_WHITESPACE_REGEX = re.compile(r'[[({] | []}),;:]')
+EXTRANEOUS_WHITESPACE_REGEX = re.compile(r'[\[({] | [\]}),;:]')
 WHITESPACE_AFTER_COMMA_REGEX = re.compile(r'[,;:]\s*(?:  |\t)')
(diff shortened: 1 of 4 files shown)


adiroiban:
This fixes the warning...but it needs more than a warning fix as Py3.7 tests fail even without the warning.

adiroiban:
I got python 3.7 installed and found the issue

```
# Python 3.7
> keyword.kwlist
['False', 'None', 'True', 'and', 'as', 'assert', 'async', 'await', 'break', 'class', 'continue', 'def', 'del', 'elif', 'else', 'except', 'finally', 'for', 'from', 'global', 'if', 'import', 'in', 'is', 'lambda', 'nonlocal', 'not', 'or', 'pass', 'raise', 'return', 'try', 'while', 'with', 'yield']

# Python 2.7
> keyword.kwlist
['and', 'as', 'assert', 'break', 'class', 'continue', 'def', 'del', 'elif', 'else', 'except', 'exec', 'finally', 'for', 'from', 'global', 'if', 'import', 'in', 'is', 'lambda', 'not', 'or', 'pass', 'print', 'raise', 'return', 'try', 'while', 'with', 'yield']
```

Now I need to see how to fix this without breaking backward compatibility... and still get consistent results.

sigmavirus24:
I don't think `async` was added in 3.7... wasn't it added in 3.5 or 3.6? It's curious that's why our tests are failing

adiroiban:
@sigmavirus24 I am a python2.7 guy... but it might be due to the order.

E271 is swapped with E272 for the case in which we have `async dev`

In python2.7 `async` was not there, so it was ignored, and the error was raised for `def`... not `async` is a keyword and is raised for `async` as it is also a keyword

adiroiban:
@sigmavirus24 I have pushed a fix for async... but maybe we should add backport all the keywords not only `print` and `async`. What do you think?

adiroiban:
@sigmavirus24 here is the keyword from python 3.5... async is not there ... this is why we did not saw this in python 3.5 

```
# Python 3.5
>>> keyword.kwlist
['False', 'None', 'True', 'and', 'as', 'assert', 'break', 'class', 'continue', 'def', 'del', 'elif', 'else', 'except', 'finally', 'for', 'from', 'global', 'if', 'import', 'in', 'is', 'lambda', 'nonlocal', 'not', 'or', 'pass', 'raise', 'return', 'try', 'while', 'with', 'yield']
```

sigmavirus24:
Thanks @adiroiban! You may want to rebase #733 :cake:

FichteFoll:
FWIW, async and await were not proper keywords in 3.5 and only valid inside an `async def` block.

gopikanand60:
I'm confused....what should I do to remove the error?

Issue #734, opened 2018-03-23
Disable 'line too long' checks just for shebang line?


jfhc:
Feature request: I would like to *not* disable E501, but ignore it just for a shebang line (if present). E.g. if the first line of the file starts with `#!`, don't check for E501, otherwise act as normal?

I posted about this [on stackoverflow](https://stackoverflow.com/q/49425288/8131703):

> So if the max line length is 5 (for sake of example), with this file:
> 
>     #!/run/stuff
>     x=3
>     print(x)
> 
> They should complain about the third line but not the first one. 
> 
> But with this file:
> 
>     x = 1 + 1 + 1
>     # Print the result
>     print(x)
> 
> It should complain about all three lines.

hoylemd:
I don't have a strong opinion either way, but I'm quite curious what shebang lines would run into the line length limit.  Is it just that the python executable's path is expecially verbose, or is there more to shebangs than I know?

sigmavirus24:
So, here's my reasoning behind asking @jfhc to create this request:

Shebang lines are special. They can be unfortunately long - I usually only see this with auto-generated lines but it's not outside the realm of possibility that it could be hand-crafted as well. Normally, my answer would be to just tack `# noqa` onto the end, but in this case, that would break the shebang. There's no good way of handling this without changing the style of the entire file/project. To ignore shebangs is incredibly easy (if first line and starts with `#!` skip line length check)

jfhc:
In my case the shebangs are long because of the way my workplace manages python environments (so my shebang line essentially points to a wrapper script with some extra arguments which grabs the right python for the script).

hoylemd:
@jfhc Oh neat! I didn't know you could do that.

Got my vote then.

sigmavirus24:
@jfhc I'd happily merge a PR that adds this feature. Would you be interested in contributing it?

jfhc:
I'd be up for giving it a go :-)

sigmavirus24:
Please do @jfhc 😃

PR #736, opened 2018-03-28
Ignore length of shebang line

jfhc:
Add check to avoid long line error on shebang lines (as discussed in [Issue 734](https://github.com/PyCQA/pycodestyle/issues/734)), and a test for this behaviour.

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -260,7 +260,8 @@ def trailing_blank_lines(physical_line, lines, line_number, total_lines):
 
 
 @register_check
-def maximum_line_length(physical_line, max_line_length, multiline, noqa):
+def maximum_line_length(physical_line, max_line_length, multiline,
+                        line_number, noqa):
     r"""Limit all lines to a maximum of 79 characters.
 
     There are still many devices around that are limited to 80 character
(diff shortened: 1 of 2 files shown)


pycodestyle.py
```diff
@@ -275,6 +276,9 @@ def maximum_line_length(physical_line, max_line_length, multiline, noqa):
     line = physical_line.rstrip()
     length = len(line)
     if length > max_line_length and not noqa:
+        # Special case: ignore long shebang lines.
+        if line_number == 1 and length > 1 and line[:2] == '#!':
```
sigmavirus24: Is `length > 1` required here? We're in the base where `length > max_line_length` and the slice you're using works even on empty strings:

```py
>>> ''[:2]
''
```

And `''` can be safely compared to `#!`

hoylemd: `length > 1` seems redundant to me, since we already know `length > max_line_length` from L280.

Also, what do you think of doing `line.startswith('#!')` instead of `line[:2] == '#!'`?  It's more characters, but seems *slightly* more readable to me.

hoylemd: Haha @sigmavirus24 [Jynx](https://github.com/PyCQA/pycodestyle/pull/736#discussion_r178100679) ;)

sigmavirus24: 12 minutes versus 8 minutes 😛

jfhc:
Good points, both - I didn't know about startswith!

I have *d this up by trying to do the edit in the GitHub web file editor thing, hold on....

jfhc:
How does the current version look?

sigmavirus24:
Looks great to me. :+1:

Issue #753, opened 2018-04-15
E252 raised on annotated expression-based defaults involving equals


joshbode:
E252 is being raised on what look to be valid annotated, expression-based arguments which contain equals.

Example:
```python
# foo.py


def _default(name: str) -> float:

    defaults = {'f': 100.0}

    return defaults.get(name, 0.0)


def f(x, y: float = _default(name='f')) -> float:
    """Test function."""

    return x + y
```

```bash
$ pycodestyle -v foo.py
user configuration: /Users/josh/.config/pycodestyle
checking foo.py
/tmp/foo.py:11:34: E252 missing whitespace around parameter equals
/tmp/foo.py:11:35: E252 missing whitespace around parameter equals

$ pycodestyle --version
2.4.0

$ cat ~/.config/pycodestyle
[pycodestyle]
max-line-length = 160
```

emcek:
@taion 
looks like commit: d9efb8f46220dc24f74cb8e0f963786d47deecdc missing something since it says:

> Don't use spaces around the '=' sign when used to indicate a
> keyword argument or a default parameter value, except when using a type annotation.

taion:
Definitely looks like my bug. Let me see.

samuelcolvin:
Any chance of a release including a fix for this? I'm coming across this more and more. Thanks.

antoine-gallix:
Yes please, a release!

PR #759, opened 2018-04-18
Fix detection of annotated argument defaults

taion:
Fixes https://github.com/PyCQA/pycodestyle/issues/753


pycodestyle.py
```diff
@@ -937,9 +937,9 @@ def whitespace_around_named_parameter_equals(logical_line, tokens):
                 parens -= 1
             elif in_def and text == ':' and parens == 1:
                 annotated_func_arg = True
-            elif parens and text == ',' and parens == 1:
```
taion: Checking both `parens` and `parens == 1` was redundant

sigmavirus24:
Thanks for touching this up @taion!

PR #777, opened 2018-07-06
Allow single element set literal

ethanhs:
I noticed recently that this was not allowed, but I feel that it should be.

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -482,11 +482,12 @@ def missing_whitespace(logical_line):
         char = line[index]
         if char in ',;:' and line[index + 1] not in WHITESPACE:
             before = line[:index]
+            end_char = line[index + 1]
             if char == ':' and before.count('[') > before.count(']') and \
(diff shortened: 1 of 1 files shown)


sigmavirus24:
There's a difference, though, one can not create a single element tuple without the comma, i.e.,

```py
(1) != (1,)
```

However,

```py
{1} == {1,}
```

One arises out of *necessity* the other is your personal preference which isn't the sole job of this tool to enforce.

FichteFoll:
What about `[1,]` then? Wouldn't that be in the same boat?

Using trailing commas is useful for multi-line collections or enumerations (like function parameters) because you can easily add a new line without confusing diff tools and because all member lines share the same punctuation. In a single line (or single-element, even) collection, you don't benefit from either because it's a single line already and the punctuation is definitely distinct through the closing bracket.

ethanhs:
Alright fine, if people don't want it, I'll close this.

PR #793, opened 2018-09-08
Add pep8 entry point to setup.py

thechief389:
For compatibility, I added a pep8 entry point.


sigmavirus24:
Thank you, for sending this along.

Unfortunately we will not merge this because we explicitly discontinued that command-line tool name. We provided a transitional period where the pep8 tool provided both command-line tool names, but we decided against doing the same with pycodestyle.

Cheers,

Issue #795, opened 2018-09-11
Checking for invalid escape sequences in docstrings yields negative column numbers


FichteFoll:
```
/tmp/test Δ cat test.py
def func():
    """this is a
    long-ish docstring
    with invalid \d escapes
    """
    pass

/tmp/test Λ pycodestyle test.py
test.py:5:-12: W605 invalid escape sequence '\d'
 ?=1
/tmp/test $ pycodestyle --version
2.4.0
```

FichteFoll:
I suspect it's [this line](https://github.com/PyCQA/pycodestyle/blob/566cdc0cb22e5530902e456d0b315403ebab980c/pycodestyle.py#L1538). I have no idea how this can return anything less than `-1`, but this clearly wasn't designed for multi-line strings.

sigmavirus24:
Worth keeping in mind that `logical_line` is different from `physical_line` for pycodestyle too and this check uses the `logical_line` that means that we consider that to all be one string on one line. I suspect using `physical_line` might present other challenges but would be worth trying to use. Unfortunately, I see no point in the future where I will have time to do that work

asottile:
This is probably due to https://bugs.python.org/issue16806 (try the same in `pypy` where this is fixed)

mgedmin:
-12, -7, bah!  I give you -3036:

```
./tests/test_safety.py:102:-3036: W605 invalid escape sequence '\_'
./tests/test_safety.py:102:-2770: W605 invalid escape sequence '\_'
./tests/test_safety.py:102:-2566: W605 invalid escape sequence '\_'
./tests/test_safety.py:102:-2548: W605 invalid escape sequence '\_'
./tests/test_safety.py:102:-2536: W605 invalid escape sequence '\_'
./tests/test_safety.py:102:-2528: W605 invalid escape sequence '\_'
./tests/test_safety.py:102:-2237: W605 invalid escape sequence '\_'
```

(flake8 on [this litle module](https://github.com/mgedmin/safety/blob/d39980d4a763fe0d6707879ce0a5b4d1f848889d/tests/test_safety.py).)

mgedmin:
It is sort-of fixed:
```
./tests/test_safety.py:74:71: W605 invalid escape sequence '\_'
./tests/test_safety.py:74:71: W605 invalid escape sequence '\_'
./tests/test_safety.py:74:71: W605 invalid escape sequence '\_'
[...]

PR #808, opened 2018-10-21
Fix line offset for 'invalid escape sequence'

### LINKED ISSUE #807: Invalid escape sequence points at incorrect line when inside a multi-line call
```python
import pytest


@pytest.mark.parametrize(  # line 4
    'x'
    ('\d', '\_'),  # line 6
)
def test(x):
    pass
```

```console
$ ~/workspace/venvflake8/bin/pycodestyle t.py 
t.py:4:2: W605 invalid escape sequence '\_'
t.py:4:2: W605 invalid escape sequence '\d'

$ ~/workspace/venvflake8/bin/pycodestyle  --version
2.4.0
```

The column offset appears to be incorrect as well 🤷‍♂️

asottile:
Resolves #807


FichteFoll:
Might also resolve #795. I can't test currently, however.

testsuite/W60.py
```diff
@@ -19,6 +19,10 @@
 regex = '''
 \.png$
 '''
+#: W605:2:5
```
sigmavirus24: Would you be willing to add line and position numbers to the other examples of this to prove it works in all cases?

asottile: done!  looks like this maybe fixes #795 as well:

```python
def func():
    """this is a
    long-ish docstring
    with invalid \d escapes
    """
    pass
```

```console
$ ./pycodestyle.py test.py 
test.py:2:5: W605 invalid escape sequence '\d'
```

asottile:
I've added offsets for the other tests, this should be ready for another round of review

PR #809, opened 2018-10-22
Add whitespace around -> annotating operator

### LINKED ISSUE #803: No missing-whitespace when annotating function return type
When running on
```
def foo(a: int)->int:
    pass
```
I expect there to be a warning on the function defenition, according to [pep8](https://www.python.org/dev/peps/pep-0008/#other-recommendations):
>

> Function annotations should use the normal rules for colons and always have spaces around the -> arrow if present. (See Function Annotations below for more about function annotations.)
> 
> Yes:
> ```
> def munge(input: AnyStr): ...
> def munge() -> AnyStr: ...
> ```
> No:
> ```
> def munge(input:AnyStr): ...
> def munge()->PosInt: ...
> ```
when running `pycodestyle` on that file, no warnings are issued

Tadaboody:
Closes #803

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -117,7 +117,7 @@ ARITHMETIC_OP = frozenset(['**', '*', '/', '//', '+', '-'])
 WS_OPTIONAL_OPERATORS = ARITHMETIC_OP.union(['^', '&', '|', '<<', '>>', '%'])
 WS_NEEDED_OPERATORS = frozenset([
     '**=', '*=', '/=', '//=', '+=', '-=', '!=', '<>', '<', '>',
-    '%=', '^=', '&=', '|=', '==', '<=', '>=', '<<=', '>>=', '='])
+    '%=', '^=', '&=', '|=', '==', '<=', '>=', '<<=', '>>=', '=', '->'])
 WHITESPACE = frozenset(' \t')
 NEWLINE = frozenset([tokenize.NL, tokenize.NEWLINE])
 SKIP_TOKENS = NEWLINE.union([tokenize.INDENT, tokenize.DEDENT])
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -115,9 +115,12 @@ KEYWORDS = frozenset(keyword.kwlist + ['print', 'async']) - SINGLETONS
 UNARY_OPERATORS = frozenset(['>>', '**', '*', '+', '-'])
 ARITHMETIC_OP = frozenset(['**', '*', '/', '//', '+', '-'])
 WS_OPTIONAL_OPERATORS = ARITHMETIC_OP.union(['^', '&', '|', '<<', '>>', '%'])
+# Warn for -> function annotation operator in py3.5+ (issue 803)
+FUNCTION_RETURN_ANNOTATION_OP = ['->'] if sys.version_info >= (3, 5) else []
 WS_NEEDED_OPERATORS = frozenset([
     '**=', '*=', '/=', '//=', '+=', '-=', '!=', '<>', '<', '>',
-    '%=', '^=', '&=', '|=', '==', '<=', '>=', '<<=', '>>=', '=', '->'])
+    '%=', '^=', '&=', '|=', '==', '<=', '>=', '<<=', '>>=', '='] +
+    FUNCTION_RETURN_ANNOTATION_OP)
 WHITESPACE = frozenset(' \t')
 NEWLINE = frozenset([tokenize.NL, tokenize.NEWLINE])
 SKIP_TOKENS = NEWLINE.union([tokenize.INDENT, tokenize.DEDENT])
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -802,7 +802,6 @@ def missing_whitespace_around_operator(logical_line, tokens):
     Okay: alpha[:-i]
 
     E225: i=i+1
-    E225: def foo(a, b)->int:
     E225: submitted +=1
     E225: x = x /2 - 1
     E225: z = x **y
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -802,6 +802,7 @@ def missing_whitespace_around_operator(logical_line, tokens):
     Okay: alpha[:-i]
 
     E225: i=i+1
+    E225: def foo(a, b)->int:
     E225: submitted +=1
     E225: x = x /2 - 1
     E225: z = x **y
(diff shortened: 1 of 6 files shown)


Tadaboody:
It seems that the tests that failed are for py2 and pypy which don't have any use for this, how can I tell the checks to ignore the new tests on these builds?

asottile:
@Tadaboody there's a `python3.py` testsuite file for python3-specific syntax: https://github.com/PyCQA/pycodestyle/blob/master/testsuite/python3.py

asottile:
Oh hmm, maybe that file works differently than the others -- let me peek

asottile:
yeah the whitespace control in the testsuite files is a little bit weird.

Here's a patch which fixes the blank line errors:

```diff
diff --git a/testsuite/python3.py b/testsuite/python3.py
index e48ee6a..64767c7 100644
--- a/testsuite/python3.py
+++ b/testsuite/python3.py
@@ -37,11 +37,9 @@ class Class:
 
     def m(self):
         xs: List[int] = []
-
 #: E225
 def bar(a, b)->int:
     pass
-
 #: Okay
 def baz(a, b) -> int:
     pass
```

(I'd push directly but I'm not actually a maintainer of pycodestyle (yet?))

Tadaboody:
Oh man thanks! I was sure I checked that but you just can't know 😛

Tadaboody:
Could a maintainer help with configuring the test suite? I really want this to be reported so [hhatto/autopep8](https://github.com/hhatto/autopep8) will be able to fix the -> formatting automatically

pycodestyle.py
```diff
@@ -802,6 +802,7 @@ def missing_whitespace_around_operator(logical_line, tokens):
     Okay: alpha[:-i]
 
     E225: i=i+1
+    E225: def foo(a, b)->int:
```
sigmavirus24: Have you tried removing this and running the test suite?

Tadaboody: Pushed it removed right now to run the full py2 testsuite

Tadaboody: Ran without it now, the problem as I see it is that the python2 suite still runs `testsuite/python3.py` as can be seen [here](https://travis-ci.org/PyCQA/pycodestyle/jobs/462447069#L511)

testsuite/python3.py
```diff
@@ -37,3 +37,9 @@ class Class:
 
     def m(self):
         xs: List[int] = []
+#: E227
+def bar(a, b)->int:
```
sigmavirus24: It's saying that it's finding E225 instead of E227. That sems like a valid test failure at this point

Tadaboody:
@sigmavirus24 from https://github.com/PyCQA/pycodestyle/pull/809/commits/d2229f665b91ff03b6f7ead5dcb44a47b0ea2dfa I think that the problem is that in py3.7 [the error reported is `E225`](https://travis-ci.org/PyCQA/pycodestyle/jobs/462448177#L541) where by https://github.com/PyCQA/pycodestyle/pull/809/commits/2fca2b68d14472827e4b79d10173881c74798cf9 s report you can see that [2.7 reports the error as `E227`](https://travis-ci.org/PyCQA/pycodestyle/jobs/462447069#L511). any idea why that might be?

sigmavirus24:
@Tadaboody https://github.com/PyCQA/pycodestyle/blob/d2229f665b91ff03b6f7ead5dcb44a47b0ea2dfa/pycodestyle.py#L844 is for arithmetic operators. https://github.com/PyCQA/pycodestyle/blob/d2229f665b91ff03b6f7ead5dcb44a47b0ea2dfa/pycodestyle.py#L870 is just for operators. I think E225 is the more correct. What if we only have `->` in https://github.com/PyCQA/pycodestyle/blob/d2229f665b91ff03b6f7ead5dcb44a47b0ea2dfa/pycodestyle.py#L118 if we're running on Python3?

Tadaboody:
@sigmavirus24 I implemented the change you proposed in [here](https://github.com/PyCQA/pycodestyle/pull/809/commits/473d6dd4cc29ccd259351c203a77cab47aa00115#diff-f7a99c5fc1f03c3939f0b630fe0c20f6R123) but then the tests for `->` keep running in versions lower than python 3.5 (the version type annotations were introduced).
I tried to match the file with the python version [here](https://github.com/PyCQA/pycodestyle/pull/809/commits/9a19db592e1dabaf24c9c65f1756e45a9ce7ce0c#diff-a7b1dd6e5a902fe631a893c1ac4c3186R172)

testsuite/support.py
```diff
@@ -169,11 +169,11 @@ def init_tests(pep8style):
     def run_tests(filename):
         """Run all the tests from a file."""
         # Skip tests meant for higher versions of python
-        version_match = re.match(r'python(\d)(\d*)\.py', filename)
-        if version_match:
-            file_version_tuple = (int(version_match[1]), int(
-                version_match[2])) if version_match[2] else (int(version_match[1]),)
-            if sys.version_info < file_version_tuple:
+        ver_match = re.match(r'python(\d)(\d*)\.py', filename)
```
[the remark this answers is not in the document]
sigmavirus24: Nevermind. I see that was code you added in the first place and I was just looking at a commit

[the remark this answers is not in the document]
Tadaboody: Sorry for the confusion. I renamed them to pass the style test - the lines were too long.
I just wanted to see if adding logic to the test collection would work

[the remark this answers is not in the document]
FichteFoll: `re.search('^...')` to be precise.

testsuite/support.py
```diff
@@ -168,6 +168,12 @@ def init_tests(pep8style):
 
     def run_tests(filename):
         """Run all the tests from a file."""
+        # Skip tests meant for higher versions of python
+        ver_match = re.search(r'python(\d)(\d)?\.py$', filename)
+        if ver_match:
+            test_against_version = tuple(int(val or 0) for val in ver_match.groups())
```
sigmavirus24: I probably gave you a poor suggestion here because of 

```
testsuite/support.py:174:80: E501:174:80 line too long (85 > 79 characters)
```

Perhaps

```suggestion
test_against_version = tuple(int(val or 0)
                             for val in ver_match.groups())
```

Tadaboody: It's my bad for not running a formatter before committing! maybe make an issue for pre-commit hooks for the repo, or add auto-formatting to the travis build. - so this doean't need to matter

Tadaboody:
Tests pass! 🎉 Thanks for all the help!

sigmavirus24:
:clap:

PR #820, opened 2018-11-24
refs W606, only NAME can follow await

wwwjfy:
`await NUMBER` or `await STRING` is not valid syntax

@jdufresne could you help on this? If I miss some cases, I'll add NUMBER/STRING to tests


jdufresne:
Can you add a test to demonstrate what was wrong with the previous implementation?

wwwjfy:
Right. Added. Thanks

PR #834, opened 2019-01-30
E402: Add "with" statement to allowed keywords

### LINKED ISSUE #833: E402: Add "with" to `allowed_try_keywords`
For certain imports, it would be nice to scope warning filters (e.g. in https://github.com/RobotLocomotion/drake/issues/10527).

In my case, I want to do something like:
```py
import warnings

with warnings.catch_warnings():
    warnings.filterwarnings(
        "ignore", category=ImportWarning,
        message="can't resolve package from __spec__")
    import meshcat
```

Without mods to 2.3.1, it throws a warning about this. If I add "with" to the allowed keywords, then it works.

Would it be possible to add this keyword?

EricCousineau-TRI:
Resolves #833

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -1067,7 +1067,8 @@ def module_imports_on_top_of_file(
             line = line[1:]
         return line and (line[0] == '"' or line[0] == "'")
 
-    allowed_keywords = ('try', 'except', 'else', 'finally', 'with')
+    allowed_keywords = (
+        'try', 'except', 'else', 'finally', 'with', 'if')
 
     if indent_level:  # Allow imports in conditional statement/function
         return
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -1067,7 +1067,7 @@ def module_imports_on_top_of_file(
             line = line[1:]
         return line and (line[0] == '"' or line[0] == "'")
 
-    allowed_try_keywords = ('try', 'except', 'else', 'finally')
+    allowed_keywords = ('try', 'except', 'else', 'finally', 'with')
 
     if indent_level:  # Allow imports in conditional statement/function
         return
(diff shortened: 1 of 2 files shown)


asottile:
Should `if` be added as well?

EricCousineau-TRI:
SGTM. Will file as a separate commit, and split it into a separate PR if it causes issues.

EricCousineau-TRI:
Done.

EricCousineau-TRI:
Ah, sorry about that! And thanks!

testsuite/E40.py
```diff
@@ -32,6 +32,13 @@
...
     print('made attempt to import foo')
+#: Okay
+if False:
+    import foo
+#: Okay
+with warnings.catch_warnings():
+    warnings.filterwarnings("ignore", DeprecationWarning)
+    import foo
 
 import bar
```
asottile: oh actually in order to demonstrate this test you need to have an `import bar` after each of the blocks that would have triggered E402

Basically need to add the `import bar` back to the test above and add `import bar` to the `if` test

EricCousineau-TRI: Er, I think I fixed it?

asottile: awesome, looks great

pycodestyle.py
```diff
@@ -1067,7 +1067,8 @@ def is_string_literal(line):
             line = line[1:]
         return line and (line[0] == '"' or line[0] == "'")
 
-    allowed_try_keywords = ('try', 'except', 'else', 'finally')
+    allowed_keywords = (
+        'try', 'except', 'else', 'finally', 'with', 'if')
```
GPHemsley: Should this also include `elif`?

EricCousineau-TRI: Done.

PR #835, opened 2019-01-30
CONTRIBUTING: Make virtualenv dir name consistent with gitignore

EricCousineau-TRI:
Had noticed this while doing #834.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,4 @@
 dist
 docs/_build
 build/
-venv/
+/venv*/
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -34,8 +34,10 @@ Next, ``cd`` to the pycodestyle repository that you cloned earlier and
 create, then activate a virtualenv::
 
     $ cd pycodestyle
-    $ virtualenv pycodestyle-venv
-    $ source pycodestyle-venv/bin/activate
+    $ virtualenv venv-pycodestyle
+    $ source venv-pycodestyle/bin/activate
+
+Note that ``venv*/`` is ignored via ``.gitignore``.
 
 Now you can install the pycodestyle requirements::
 


CONTRIBUTING.rst
```diff
@@ -34,8 +34,10 @@ Next, ``cd`` to the pycodestyle repository that you cloned earlier and
 create, then activate a virtualenv::
 
     $ cd pycodestyle
-    $ virtualenv pycodestyle-venv
-    $ source pycodestyle-venv/bin/activate
+    $ virtualenv venv
+    $ source venv/bin/activate
+
+Note that ``venv/`` is ignored via ``.gitignore``.
```
asottile: so I assume that the original author intended for the project name to be in the venv name so it would show up in the `PS1` -- I think we can make _both_ work with something this:

### in `.gitignore`

```
/venv*
```

### in docs

suggest `venv-pycodestyle`

(and it'll also work for the common workflow of a venv just named `venv` or `venv2` or `venv3`)

EricCousineau-TRI: Done. (Would still be nice to call out `gitignore`, 'cause that's the first thing I'll look at when generating auxiliary artifacts.)

.gitignore
```diff
@@ -5,4 +5,4 @@
 dist
 docs/_build
 build/
-venv/
+/venv*/
```
sigmavirus24: Is the leading slash correct here? Should it just be `venv*/` This isn't a regular expression, it's a glob (I thought)

EricCousineau-TRI: The leading slash in `.gitignore` implies that it should only match `venv*/` directories that directly neighbor the `.gitignore` file.

e.g. it will ignore `pycodestyle/venv-biscuit/`, but not `pycodestyle/docs/venv-cookie`.

EricCousineau-TRI: Citing da sources:
https://git-scm.com/docs/gitignore#_pattern_format

> A leading slash matches the beginning of the pathname. For example, "/*.c" matches "cat-file.c" but not "mozilla-sha1/sha1.c".

EricCousineau-TRI:
Er, and is there an easy way to retrigger Travis? Or should I just close + reopen the PR?
The failures in CPython3.5 + PyPy3.5 seem like download flakes?

sigmavirus24:
> The failures in CPython3.5 + PyPy3.5 seem like download flakes?

I've triggered both to re-build. The CPython 3.5 build at the bottom, however, is for flake8. That one is likely not related to flakey tests.

asottile:
thanks!

PR #837, opened 2019-02-01
Expect lines to be indented 8 places when tabs are used

### LINKED ISSUE #836: E117 being reported for files containing tabs
The change in https://github.com/PyCQA/pycodestyle/pull/705 released in 2.5.0 has started causing these errors for files with tabs. cc @jdufresne @sigmavirus24

jdufresne:
Fixes #836

@sciyoshi Are you able to test if this fixes the issue you reported?

### DIFF
--- a/testsuite/test_api.py
+++ b/testsuite/test_api.py
@@ -125,7 +125,7 @@ class APITestCase(unittest.TestCase):
         report = pycodestyle.StyleGuide().check_files([E11])
         stdout = sys.stdout.getvalue().splitlines()
         self.assertEqual(len(stdout), report.total_errors)
-        self.assertEqual(report.total_errors, 20)
+        self.assertEqual(report.total_errors, 24)
         self.assertFalse(sys.stderr)
         self.reset()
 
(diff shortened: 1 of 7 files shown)


testsuite/test_api.py
```diff
@@ -125,15 +125,15 @@ def test_styleguide(self):
         report = pycodestyle.StyleGuide().check_files([E11])
         stdout = sys.stdout.getvalue().splitlines()
         self.assertEqual(len(stdout), report.total_errors)
-        self.assertEqual(report.total_errors, 20)
+        self.assertEqual(report.total_errors, 24)
```
jdufresne: The new errors are:

```
pycodestyle/testsuite/E11.py:41:1: E302 expected 2 blank lines, found 0
pycodestyle/testsuite/E11.py:42:1: E101 indentation contains mixed spaces and tabs
pycodestyle/testsuite/E11.py:42:1: W191 indentation contains tabs
pycodestyle/testsuite/E11.py:42:3: E117 over-indented
```

GPHemsley:
PEP8 says:
> Use 4 spaces per indentation level.

If a tab is used as indentation (my preference), it should be assumed to have a tabwidth of 4 by default.

If other codebases are using other tabwidths (8 and 2 are probably the most common alternatives), then it might make more sense to add a config option to allow customizing tabwidth rather than having the default divert that far from PEP8.

(Note also the existence of [`flake8-tabs`](https://pypi.org/project/flake8-tabs/).)

PR #847, opened 2019-02-28
E225 Check for space around boolean operators

adamchainz:
This was documented in the docstring for `missing_whitespace_around_operator` but not implemented, allowing e.g. `1and 0` to pass. Fixed by adding test case then modifying the check to search for the 'and' and 'or' operators as well.


testsuite/E22.py
```diff
@@ -76,6 +76,10 @@
 i=i+ 1
 #: E225 E225
 i=i +1
+#: E225
+i = 1and 1
+#: E225
+i = 1or 0
```
asottile: here's a few more:

```python
1is 1
1not in []
1in []
```

adamchainz: Added support for 'is' and 'not', updating for the two-word 'not in' is beyond me right now (07:30 after a long haul flight...)

asottile: missed `not` ;)

adamchainz:
Thanks @asottile for review and @sigmavirus24 for merge!

sigmavirus24:
Thanks for the fix! When you're less jet-lagged and feel like puzzling out `not in`, feel free to send another PR.

PR #869, opened 2019-05-30
Fix #856

### LINKED ISSUE #856: E501 Combining diacritics counted against line length
It looks like unicode combining diacritics are counted as full characters, which causes `E501` to fire, even when it is visually less than 79 characters:

```python
text = ['ся́ду', 'твёрдый', 'ю́бкой', 'ба́юшки', 'опя́ть', 'съе́сть', 'жи́ть', 'éли']
a = 'а́́́́́́́́́́́́́́́́́́́'
```

It seems to me that combining characters should be excluded from the count.

reynoldsnlp:
I also alphabetized the import statements by moving `bisect` to the top.

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -49,6 +49,7 @@ W warnings
 """
 from __future__ import with_statement
 
+import bisect
 import inspect
 import keyword
 import os
@@ -56,8 +57,8 @@ import re
 import sys
 import time
 import tokenize
+import unicodedata
 import warnings
-import bisect
 
 try:
     from functools import lru_cache
@@ -283,7 +284,8 @@ def maximum_line_length(physical_line, max_line_length, multiline,
     Reports error E501.
     """
     line = physical_line.rstrip()
-    length = len(line)
+    # compute length ignoring combining diacritics
+    length = sum(not unicodedata.combining(c) for c in line)
     if length > max_line_length and not noqa:
         # Special case: ignore long shebang lines.
         if line_number == 1 and line.startswith('#!'):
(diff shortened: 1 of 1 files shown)


asottile:
Can you write a test demonstrating this change (and so it doesn't regress) -- thanks for your contribution! :tada:

reynoldsnlp:
@asottile I'm not sure I wrote that test correctly, so make sure you think it looks good.

pycodestyle.py
```diff
@@ -296,12 +297,16 @@ def maximum_line_length(physical_line, max_line_length, multiline,
...
+        # Special case: multi-byte chars and combining diacritics
+        try:
+            length = sum(not unicodedata.combining(c) for c in line)
+        except TypeError:  # Python 2 str
+            if hasattr(line, 'decode'):  # Python 2 str
+                try:
+                    length = sum(not unicodedata.combining(c)
+                                 for c in line.decode('utf-8'))
+                except UnicodeError:
+                    pass
```
asottile: I don't think we should be guessing about the types here with `hasattr` and `TypeError` -- let's be explicit and use `sys.version_info >= (3,)` and such

I'd write it like this:

```python
if sys.version_info >= (3,):
    line_text = line
else:
    try:
        line_text = line.decode('UTF-8')
    except UnicodeDecodeError:
        line_text = u''
    length = sum(not unicodedata.combining(c) for c in line)
```

reynoldsnlp: I think that a big improvement.

reynoldsnlp:
Thanks for your help! I'll be very happy to have that fixed.

reynoldsnlp:
Do the checks need to be restarted?

sigmavirus24:
Can you rebase? I expect the checks will run correctly then if you rebase and force-push

Issue #57, opened 2012-04-30
Avoid Extraneous Whitespace Around Keywords


rconradharris:
The current code catches whitespace issues around operators but not around keywords.

The proposal is to add a new error-class, 27X which would contain whitespace errors related to keywords. The implementation would be done in a very smilar way to operators.

PR #870, opened 2019-06-11
Allow spaces before lambda.

anntzer:
Currently, pycodestyle allows
```
{
    "key":        1,
    "longer-key": 2,
}
```
(actually it emits a E241 "multiple spaces after ":", but that's
suppressed by default) but disallows
```
{
    "key":        lambda: 1,
    "longer-key": lambda: 2,
}
```
because the first entry has "multiple spaces before keyword" (before the
lambda), even though the cases are clearly similar.

Fix that by excluding lambda from that check.

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -141,7 +141,7 @@ COMPARE_SINGLETON_REGEX = re.compile(r'(\bNone|\bFalse|\bTrue)?\s*([=!]=)'
 COMPARE_NEGATIVE_REGEX = re.compile(r'\b(not)\s+[^][)(}{ ]+\s+(in|is)\s')
 COMPARE_TYPE_REGEX = re.compile(r'(?:[=!]=|is(?:\s+not)?)\s+type(?:s.\w+Type'
                                 r'|\s*\(\s*([^)]*[^ )])\s*\))')
-KEYWORD_REGEX = re.compile(r'(\s*)\b(?:%s)\b(\s*)' % r'|'.join(KEYWORDS))
+KEYWORD_REGEX = re.compile(r'(\s*)\b(%s)\b(\s*)' % r'|'.join(KEYWORDS))
 OPERATOR_REGEX = re.compile(r'(?:[^,\s])(\s*)(?:[-+*/|!<=>%&^]+)(\s*)')
 LAMBDA_REGEX = re.compile(r'\blambda\b')
 HUNK_REGEX = re.compile(r'^@@ -\d+(?:,\d+)? \+(\d+)(?:,(\d+))? @@.*$')
@@ -447,17 +447,17 @@ def whitespace_around_keywords(logical_line):
     E274: True\tand False
     """
     for match in KEYWORD_REGEX.finditer(logical_line):
-        before, after = match.groups()
+        before, kw, after = match.groups()
 
-        if '\t' in before:
+        if '\t' in before and kw != "lambda":
             yield match.start(1), "E274 tab before keyword"
-        elif len(before) > 1:
+        elif len(before) > 1 and kw != "lambda":
             yield match.start(1), "E272 multiple spaces before keyword"
 
         if '\t' in after:
-            yield match.start(2), "E273 tab after keyword"
+            yield match.start(3), "E273 tab after keyword"
         elif len(after) > 1:
-            yield match.start(2), "E271 multiple spaces after keyword"
+            yield match.start(3), "E271 multiple spaces after keyword"
 
 
 @register_check
--- a/testsuite/E27.py
+++ b/testsuite/E27.py
@@ -42,3 +42,8 @@ try:
     from importable.module import(e, f)
 except ImportError:
     pass
+#: E241
+{  # (but no E272)
+    "key":        lambda: None,
+    "longer-key": lambda: None,
+}


pycodestyle.py
```diff
@@ -447,11 +447,11 @@ def whitespace_around_keywords(logical_line):
     E274: True\tand False
     """
     for match in KEYWORD_REGEX.finditer(logical_line):
-        before, after = match.groups()
+        before, kw, after = match.groups()
```
FichteFoll: I suggest using the more explicit `keyword` instead of `kw` for this variable name.

anntzer: That would shadow the global "keyword" module name.

FichteFoll: Ah, that's unfortunate. I missed this due to only reading the diff.

anntzer:
Kindly bumping?

asottile:
I believe this regresses:

```python
x =                 lambda: 1
```

and doesn't address

```python
{
    'x':      not 4,
    'y':     print(6),
}
```

I can't think of a correct heuristic to use here without breaking all of those so I'm not sure this PR is possible

anntzer:
If you look at #57 which is the origin of E272, I think the intent was clearly to cover "binary" keywords (`and`, `or`) rather than unary ones (`lambda`) -- certainly the spaces before `and` in `a  and b` and the spaces before `not` in `{a:  not b}` don't play the same role.  Likewise if your preferred style guide allows
```
{
    'x':    1,
    'long': 2,
}
```
then it doesn't make sense to disallow
```
{
    'x':    lambda: 1,
    'long': lambda: 2,
}
```
so I guess a solution would be to change E272 to only cover binary keywords?

asottile:
I don't think that's correct, the intent appears to be to make it the same as operator whitespace checks -- which if we try the same for unary operators:

```console
$ ~/opt/venv/bin/pycodestyle t.py 
t.py:2:9: E221 multiple spaces before operator
$ cat t.py
x = {
    'a':  +5,
}
```

anntzer:
Well, that's not what the docstring in https://github.com/PyCQA/pycodestyle/pull/58/files#diff-429d64b260de27fe3688921a1592b3f6R343 suggests.

Also consider that with
```
x =  +5
y =  3
{'z':  +5}
{'t':  3}
```
pycodestyle currently gives
```
/tmp/test.py:1:4: E222 multiple spaces after operator
/tmp/test.py:2:4: E222 multiple spaces after operator
/tmp/test.py:3:6: E221 multiple spaces before operator
```
which is a bit inconsistent? (also note that E222 is technically worded incorrectly because `=` is not an "operator" (this term is defined at https://docs.python.org/3/reference/lexical_analysis.html#operators and https://docs.python.org/3/reference/expressions.html#operator-precedence) -- anyways that's a side point)

I guess for full consistency I would also suggest to change E221?  Again if your style guide allows spaces to align after colons
```
{
    'x':    42,
    'long': 123,
}
```
it seems strange that you'd need to add a new exception when you introduce a negative number
```
{
    'x':    42,
    'long': 123,
    'y':    -42,
}
```
whereas there are plenty of other unrelated places where you may still want to reject spaces before a unary minus.

asottile:
right, my assertion is that the docstring (and lack of tests) was an oversight and the behaviour is correct

anntzer:
What is your opinion on the examples I give about spaces after colons?

asottile:
it seems to me that those should act *more* like E222 and not *less* like it -- I would hope there would be an error for both lines

anntzer:
E222 is about operators and just like for `=`, `:` is not an operator but a delimiter.  Instead these *do* emit E241 (multiple spaces after ':'), but that's silenced by default.

asottile:
the tokenizer treats them both as operators:

```console
$ python3 -m tokenize <<< 'x = 5'
1,0-1,1:            NAME           'x'            
1,2-1,3:            OP             '='            
1,4-1,5:            NUMBER         '5'            
1,5-1,6:            NEWLINE        '\n'           
2,0-2,0:            ENDMARKER      ''             
$ python3 -m tokenize <<< '{"x": 5}'
1,0-1,1:            OP             '{'            
1,1-1,4:            STRING         '"x"'          
1,4-1,5:            OP             ':'            
1,6-1,7:            NUMBER         '5'            
1,7-1,8:            OP             '}'            
1,8-1,9:            NEWLINE        '\n'           
2,0-2,0:            ENDMARKER      ''  
```

anntzer:
https://docs.python.org/3/library/tokenize.html#module-tokenize says

> To simplify token stream handling, all operator and delimiter tokens and Ellipsis are returned using the generic OP token type. The exact type can be determined by checking the exact_type property on the named tuple returned from tokenize.tokenize().

so it's called OP but it's *not* an operator token, despite the name.

asottile:
seems like you're splitting hairs a bit -- an assignment statement involves an assignment operator (`=`) (the grammar may call it a delimeter, but the docs refer to "assignment operator" and even "walrus operator" for assignment expressions)

to me, the `:` in a dictionary is an assignment operator, the left hand side key is being assigned the value from the right hand side (the grammar again may call this a delimiter)

I think a more amenable proposal would be to shift the assignments to a separate code such that you can ignore it.  I think this patch is currently unacceptable as it allows ugly code that was previously prevented

I don't really wish to discuss this further, I really don't have any say on this project -- I'm just an interested bystander and would be sad to see the linter weakened for a style preference that to me is incorrect.

thanks for your time

asottile:
(I'm unsubscribing from this PR -- if you'd like to address me directly please @-mention me -- thanks!)

anntzer:
`:=` and `=` are obviously not the same (`foo := bar` is an expression, `foo = bar` is not).
In any case my point is that whether you allow spaces in `foo     + bar` and whether you allow spaces in `'x':    +bar` seem unrelated enough to warrant separate codes.

anntzer:
Closing as inactive and due to the disagreement about the design here.  Feel free to reuse the patch if desired.

Issue #867, opened 2019-05-23
Support PEP 570


sproshev:
https://www.python.org/dev/peps/pep-0570

E225 should be disabled for `/` in parameters list.

asottile:
for a concrete testcase:

```python
def f(a, b, /, c, d, *, e, f):
    pass
```

currently failing with:

```console
$ python t.py
$ pycodestyle t.py 
t.py:1:14: E225 missing whitespace around operator
```

Nekit10:
I think the issue can be closed now

jirassimok:
There's still one more case to cover: a function with only positional-only parameters still has the same problem, with a missing space before the closing parenthesis.

These are the additional test cases (for `testsuite/python38.py`) that should be passed before this issue is closed:

```
#: Okay
def f(a, b, /):
    pass
#: E225:1:7
def f(/, a, b):
    pass

```

I have verified that you can't use the exact same fix for this as for other positional-only parameters.

asottile:
@jirassimok can confirm, both of those shouldn't cause errors and currently do

curiously, my patch suggestion here seems to fix this: https://github.com/PyCQA/pycodestyle/pull/872#issuecomment-543396953

jirassimok:
I've been looking into it a bit. The problem occurs for two reasons.

First, the block starting at [line 862](https://github.com/PyCQA/pycodestyle/blob/d69c15eb7ecf77e94988fb55207a78936b48079c/pycodestyle.py#L896) tries to ensure that there is equal whitespace on either side of the slash. Then in a future iteration of the loop, it finds the space missing and issues E225.

Clearing that requirement, as I did in b5e89680693d443aa3882f1f7318191e00b12201, does get rid of the problem, but it also stops the error if the slash is given as the first parameter, which is illegal.

@asottile, your patch works because making the slash a unary operator lets it be handled by the block starting at [line 886](https://github.com/PyCQA/pycodestyle/blob/d69c15eb7ecf77e94988fb55207a78936b48079c/pycodestyle.py#L886), which does basically the same thing as my commit (though it has another condition in it).

But on second thought, I think that the last test case I added is actually incorrect. This is a style checker, not a syntax checker, so it's all right if it doesn't complain about a parameter list starting with a slash.

jirassimok:
@asottile I agreed with the concerns others expressed on #872, but I now believe your solution is better.

[...]

PR #872, opened 2019-07-01
Added support for PEP 570 (#867)

Nekit10:
See Issue #867 or https://www.python.org/dev/peps/pep-0570/


asottile:
fwiw, this patch also fixes the problem, I don't know which one is more correct (there other "unary operators" covering the splat / etc. operations in this list (it's how named-only arguments aren't flagged):

```diff
diff --git a/pycodestyle.py b/pycodestyle.py
index 6ad5456..a0c2898 100755
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -112,7 +112,7 @@ REPORT_FORMAT = {
 PyCF_ONLY_AST = 1024
 SINGLETONS = frozenset(['False', 'None', 'True'])
 KEYWORDS = frozenset(keyword.kwlist + ['print', 'async']) - SINGLETONS
-UNARY_OPERATORS = frozenset(['>>', '**', '*', '+', '-'])
+UNARY_OPERATORS = frozenset(['>>', '**', '*', '/', '+', '-'])
 ARITHMETIC_OP = frozenset(['**', '*', '/', '//', '+', '-'])
 WS_OPTIONAL_OPERATORS = ARITHMETIC_OP.union(['^', '&', '|', '<<', '>>', '%'])
 # Warn for -> function annotation operator in py3.5+ (issue 803)
diff --git a/testsuite/python38.py b/testsuite/python38.py
new file mode 100644
index 0000000..2132cd5
--- /dev/null
+++ b/testsuite/python38.py
@@ -0,0 +1,2 @@
+def f(a, /, b):
+    pass
```

mikeholler:
Any updates on this? I'm having issues as well.

aviramha:
> code seems fine, can you make `testsuite/python38.py` with:
> 
> ```python
> def f(a, /, b):
>     pass
> ```

Seems author is stale, should I/anyone else just create a new PR with a test added?

Nekit10:
I will add tests tonight

aviramha:
Any update?

testsuite/python38
```diff
@@ -0,0 +1,3 @@
+#: Okay
```
asottile: needs to be `python38.py`

pycodestyle.py
```diff
@@ -855,6 +855,10 @@ def missing_whitespace_around_operator(logical_line, tokens):
                 # Tolerate the "<>" operator, even if running Python 3
                 # Deal with Python 3's annotated return value "->"
                 pass
+            elif prev_text == '/' and text == ',':
+                # Tolerate the "/" operator in function definition
+                # For more info see PEP570source
```
asottile:
```suggestion
                # For more info see PEP570
```

asottile:
actually, looks like you'll need to rebase now since `python38.py` got created on master

Nekit10:
> this seems fine to me, a smaller patch however:
> 
> ```diff
> -UNARY_OPERATORS = frozenset(['>>', '**', '*', '+', '-'])
> +UNARY_OPERATORS = frozenset(['>>', '**', '*', '/', '+', '-'])
> ```
> 
> (this is how keyword-only arguments are handled currently)
> 
> 
I think it is not a good idea 'cause in this case '/' is not an unary operator

asottile:
> > this seems fine to me, a smaller patch however:
> > ```diff
> > -UNARY_OPERATORS = frozenset(['>>', '**', '*', '+', '-'])
> > +UNARY_OPERATORS = frozenset(['>>', '**', '*', '/', '+', '-'])
> > ```
> > 
> > 
> > (this is how keyword-only arguments are handled currently)
> I think it is not a good idea 'cause in this case '/' is not an unary operator

neither is `>>` nor `*` nor `**`

FichteFoll:
`*` and `**` kinda are unary for unpacking, but they aren't operators in that case. I have no idea why `>>` is in that list, however.

Either way, since * and ** are allowed for unpacking in quite a few situations, it kind of makes sense to have them in that list, but `/` is really only valid in the parameter list and the suggested patch is more targeted and thus makes more sense to me.

asottile:
`>>` is probably for python 2

charisz:
Waiting for release

PR #878, opened 2019-07-28
Docstring may also be prefixed with U and R.

xuhdev:


FichteFoll:
Also just `f`.

pycodestyle.py
```diff
@@ -133,7 +133,9 @@ def lru_cache(maxsize=128):  # noqa as it's a fake implementation.
 RAISE_COMMA_REGEX = re.compile(r'raise\s+\w+\s*,')
 RERAISE_COMMA_REGEX = re.compile(r'raise\s+\w+\s*,.*,\s*\w+\s*$')
 ERRORCODE_REGEX = re.compile(r'\b[A-Z]\d{3}\b')
-DOCSTRING_REGEX = re.compile(r'u?r?["\']')
+DOCSTRING_REGEX = re.compile('(u|r|ur|ru' +
+                             ('|fr|rf' if sys.version_info >= (3, 6) else '') +
```
asottile: docstrings cannot be prefixed with `f`:

```pycon
>>> class C:
...     f'hi'
... 
>>> C.__doc__
>>> 
```

xuhdev: Thanks, I updated the PR and description to drop the change regarding `f` and `F`.

pycodestyle.py
```diff
@@ -133,7 +133,7 @@ def lru_cache(maxsize=128):  # noqa as it's a fake implementation.
 RAISE_COMMA_REGEX = re.compile(r'raise\s+\w+\s*,')
 RERAISE_COMMA_REGEX = re.compile(r'raise\s+\w+\s*,.*,\s*\w+\s*$')
 ERRORCODE_REGEX = re.compile(r'\b[A-Z]\d{3}\b')
-DOCSTRING_REGEX = re.compile(r'u?r?["\']')
```
asottile: russian strings (`ru'foo'`) are not valid -- could you also add a test case to demonstrate your change while you're at it -- thanks!


```console
$ python2
Python 2.7.15+ (default, Nov 27 2018, 23:36:35) 
[GCC 7.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> ru'foo'
  File "<stdin>", line 1
    ru'foo'
          ^
SyntaxError: invalid syntax
>>> 
$ python3
Python 3.6.8 (default, Jan 14 2019, 11:02:34) 
[GCC 8.0.1 20180414 (experimental) [trunk revision 259383]] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> ru'foo'
  File "<stdin>", line 1
    ru'foo'
          ^
SyntaxError: invalid syntax
```

xuhdev:
I updated the PR. I'm not sure how to add relevant test though; pycodestyle does not appear to check specifically for string prefixes (perhaps this can be added as a feature?).

asottile:
> I updated the PR. I'm not sure how to add relevant test though; pycodestyle does not appear to check specifically for string prefixes.

I'm not sure what this PR fixes then -- did you have a particular case where this fixed something?

I was able to debug what this is intended to protect against (I think?)

```diff
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -374,6 +374,9 @@ def blank_lines(logical_line, blank_lines, indent_level, line_number,
                     previous_indent_level < indent_level or
                     DOCSTRING_REGEX.match(previous_logical)
                     ):
+
+                if not DOCSTRING_REGEX.match(previous_logical):
+                    breakpoint()
                 ancestor_level = indent_level
                 nested = False
                 # Search backwards for a def ancestor or tree root
```

this is hit for the following testcase from the testsuite:

```python
class X:

    def a():
        pass
    def b():
        pass
```

Adjusting this slightly:

```python
class X:

    def a():
        """docstring"""
    def b():
        pass
```

that passes with no errors:

```console
$ python3.7 -m pycodestyle t.py
$
```

But, if we adjust it:

```python
class X:

    def a():
        U"""docstring"""
    def b():
        pass
```

```console
$ python3.7 -m pycodestyle t.py
t.py:5:5: E301 expected 1 blank line, found 0
```

You can probably use this difference to write a testcase I imagine?

pycodestyle.py
sigmavirus24: @xuhdev I'm looking at the PR diff here and am not seeing a test or change that matches what @asottile asked for

PR #879, opened 2019-07-31
Add support for assignment expressions

### LINKED ISSUE #865: Support PEP 572
https://www.python.org/dev/peps/pep-0572/

E203 and E231 should be disabled for `:=` in assignment expressions.

FichteFoll:
Introduced in Python 3.8 with PEP-572. Refer to https://www.python.org/dev/peps/pep-0572.

I took some tests from the PEP's examples section.

Closes #865.

(Really not a fan of binary operators before line breaks.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,8 @@ matrix:
       env: TOXENV=py36
     - python: 3.7
       env: TOXENV=py37
+    - python: 3.8
+      env: TOXENV=py38
     - python: pypy2.7-6.0
       env: TOXENV=pypy
     - python: pypy3.5-6.0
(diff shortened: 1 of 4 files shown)


FichteFoll:
Since Python 3.8 has been out for almost a week now, a new release with support for assignment expressions and the other syntax addition (#872) should probably be provided sooner rather than later. Without this change, pycodestyle will remain oblivious to the new syntax and issue multiple complaints, unless you disable checking for the lines that use it.

.travis.yml
```diff
@@ -17,6 +17,8 @@ matrix:
       env: TOXENV=py36
     - python: 3.7
       env: TOXENV=py37
+    - python: 3.8-dev
```
sigmavirus24: Given that 3.8 has been released now, can we update this yet on Travis?

FichteFoll: It's included in [this list](https://docs.travis-ci.com/user/languages/python/#specifying-python-versions), so let's try it out.

FichteFoll: It's included in [this list](https://docs.travis-ci.com/user/languages/python/#specifying-python-versions), so let's try it out.

fr0der1c:
Is there a planned release date, please?

BornForFever:
> Thanks for the patch!

So when will it be updated? It's been almost half a year since Python 3.8 was released!

PR #881, opened 2019-08-12
Structured reporter

mjbommar:
Allow users to easily access structured output from `Checker` directly by providing a simple Report class with a list of dictionaries.

See example below:
```
>>> report = pycodestyle.ListReport(options=style_guide.options)
>>> checker = pycodestyle.Checker(lines=lines, report=report)
>>> checker.check_all()
>>> print(report.entries)
[{'message_class': 'error',
   'message_code': 'W292',
   'filename': 'stdin',
   'lineno': 1,
   'offset': 57,
   'message': 'no newline at end of file',
   'source': 'from .client import Client, ResultError, JobNotFoundError'},
  {'message_class': 'error',
   'message_code': 'E501',
   'filename': 'stdin',
   'lineno': 177,
   'offset': 79,
   'message': 'line too long (105 > 79 characters)',
   'source': '        # we can terminate quickly by checking if the process is not running and it has no queued output.'},
  {'message_class': 'error',
   'message_code': 'E122',
   'filename': 'stdin',
   'lineno': 2,
   'offset': 0,
   'message': 'continuation line missing indentation or outdented',
   'source': 'import uuid'},
...
```

See also discussion at: https://github.com/PyCQA/pyflakes/pull/462

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -2321,6 +2321,104 @@ class DiffReport(StandardReport):
         return super(DiffReport, self).error(line_number, offset, text, check)
 
 
+class ListReport(object):
+    """Collect the results of the checks into a structured
+    list of dictionary entries."""
+
+    def __init__(self, options):
+        self._benchmark_keys = options.benchmark_keys
+        self._ignore_code = options.ignore_code
+        # Results
+        self.elapsed = 0
+        self.total_errors = 0
+        self.counters = dict.fromkeys(self._benchmark_keys, 0)
+        self.messages = {}
+        self.entries = []
+
+    def start(self):
+        """Start the timer."""
+        self._start_time = time.time()
+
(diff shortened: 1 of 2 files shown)


sigmavirus24:
Hi again,

Even if we were to accept a structured output format for pycodestyle, we wouldn't accept Python REPRs. Many people use the PyCQA suite of tools from other languages to use and build continuous integration pipelines. Python REPRs are not portable across Java, Perl, Ruby, etc.

Further, as @asottile pointed out on PyFlakes, there's already a plugin to Flake8 that does this.

Finally, this doesn't manifest in usable way by anyone who isn't writing their own orchestration with pycodestyle using the Python API directly. As such, if it were to be accepted, it'd immediately be dead code and very likely to just disappear.

Issue #830, opened 2019-01-27
E721: Naive for modules named `types`


coxley:
This relates to #671 

## Install Method and Version
```
pip install --user pycodestyle
pycodestyle --version
2.4.0
```

## Context

E721 is the error for comparing the `type(x)` of something with another instead of using `isinstance`. Looking at #671, I thought this was something that I could submit a PR for. [Looking at the testsuite for E72, however](https://github.com/PyCQA/pycodestyle/blob/master/testsuite/E72.py#L15,L16), this is expected behavior currently.

## Problem

Any object named `types` is assumed to be the stdlib `types` module. There are many just reasons for a library to have a `types` module making selective disabling of E721 mandatory. (or alias imports).

```python

from foo import types

some_value = get_column_from_db()
if some_value == types.MyThing.SOME_PROPERTY:
    ...

```

Submitting an issue instead of PR because I'm not sure what context can be available to checks. :) The goal of the project is to be quick so the `ast` is out of the question. With that, either:

* The scope of the check should be reduced to not include `types` module checking
  * If we can't do it right, we shouldn't settle for "good enough" for an E-level check
* Regex could be expanded to include known contents of `types`
  * This still seems off because more maintenance

coxley:
@froody: is this fixed and deployed in a recent release?

asottile:
@coxley no, the PR which mentioned this was closed without merge

coxley:
Ah whoops, missed that. Thanks.

coxley:
Ping ping ping. We run into so many false-positives with this.

asottile:
@coxley please don't ping issues like that, I'd suggest attempting a patch

asfaltboy:
I've got the same issue in one of my modules, importing a local `types` module and comparing some `types.MyEnum` members returns a E721. But, I failed to reproduce a simplified isolated example; so there must be something else involved here... I'll try to break into the code and see what happens.

Issue #671, opened 2017-07-31
E721 false positives


saleemrashid:
In https://github.com/trezor/python-trezor/pull/124, there are comparisons like `msg.type == types.PinMatrixRequestType_Current`, where `msg.type` and `types.PinMatrixRequestType_Current` are `int`.

 This gives a false positive because `COMPARE_TYPE_REGEX` does not check if `type` is a function call, and `PinMatrixRequestType_Current` contains the word `Type`.

asottile:
while #830 is a duplicate of this one, I believe that one has more information so we can continue discussion there!

PR #883, opened 2019-09-05
Fix COMPARE_TYPE_REGEX to handle alternative 'types' modules

froody:
https://github.com/PyCQA/pycodestyle/issues/671
https://github.com/PyCQA/pycodestyle/issues/830


froody:
need to fix tests

PR #912, opened 2020-02-25
Support visual indent of continuation lines after with/assert/raise.

### LINKED ISSUE #569: E127 on slightly modified PEP-8 approved example (multiline with)
I took the multiline with statement from PEP-8 and tested a few variations. I only get an error when the statement is inside a function and preceded by another statement. This seems related to #316 and #256. git bisect confirms: the first bad commit is 974748fd9a0ecdd494cdc51aeb16c54d38256d58.
## Steps to reproduce

Run pycodestyle on:

```
def f1():
    print('foo')
    with open('/path/to/some/file/you/want/to/read') as file_1, \
         open('/path/to/some/file/being/written', 'w') as file_2:
        file_2.write(file_1.read())


def f2():
    with open('/path/to/some/file/you/want/to/read') as file_1, \
         open('/path/to/some/file/being/written', 'w') as file_2:
        file_2.write(file_1.read())

print('foo')
with open('/path/to/some/file/you/want/to/read') as file_1, \
     open('/path/to/some/file/being/written', 'w') as file_2:
    file_2.write(file_1.read())
```
## Result

`test.py:4:10: E127 continuation line over-indented for visual indent`
## Expected

No error, or at least the same behavior between all three versions. Before 974748fd9a0ecdd494cdc51aeb16c54d38256d58, pycodestyle reported E126 errors for the other two variations.

anntzer:
"with" is likely the most common case, and this indentation is
explicitly given as example by PEP8 (under "maximum line length").

Closes #569.

### DIFF
--- a/testsuite/E12.py
+++ b/testsuite/E12.py
@@ -372,4 +372,26 @@ print(True)
 
 print(a
 , end=' ')
-#:
+#: E127:4:12
+def foo():
+    pass
+    raise 123 + \
+           123
+#: E127:4:13
+class Eggs:
+    pass
+    assert 123456 == \
+            123456
+#: E127:4:11
+def f1():
+    print('foo')
+    with open('/path/to/some/file/you/want/to/read') as file_1, \
+          open('/path/to/some/file/being/written', 'w') as file_2:
+        file_2.write(file_1.read())
+#: E127:5:11
+def f1():
+    print('foo')
+    with open('/path/to/some/file/you/want/to/read') as file_1, \
+         open('/path/to/some/file/being/written', 'w') as file_2, \
+          open('later-misindent'):
+        file_2.write(file_1.read())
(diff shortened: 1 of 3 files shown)


testsuite/E12.py
```diff
@@ -373,3 +373,18 @@ def example_issue254():
...
+class Eggs:
+    pass
+    assert 123456 == \
+            123456
+#:
+def f1():
+    print('foo')
+    with open('/path/to/some/file/you/want/to/read') as file_1, \
+          open('/path/to/some/file/being/written', 'w') as file_2:
+        file_2.write(file_1.read())
```
asottile: none of these three align properly -- they should produce an error I believe?

anntzer: they do.
not sure I understand the syntax of the test file actually... I added `#: E127:4:...` and looks like tests pass
(the error is produced when manually testing as well)

asottile: :confused: weird!  I'll look into why that is and fix that up :+1:

FichteFoll:
Does it work for more than one line continuation as well?

anntzer:
Yes, added tests for that.

PR #913, opened 2020-02-25
Correctly allow *two* blank lines after a block of one-liners.

anntzer:
Previously
```
def oneliner(): pass
def otheroneliner(): pass

def really_long_func():
    with_some_contents
```
would raise an "E302: expected 2 blank lines, found zero" at the last
line of the one liner.  Ultimately, this is due to `expand_indent` being
passed a line whose contents are just a newline and nothing else, and
`expand_indent` thinking that the line is indented by 1 character (the
newline), which is wrong.

Fix that by just stripping the newline, and modify a test to cover this
case.

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -1794,6 +1794,7 @@ def expand_indent(line):
     >>> expand_indent('        \t')
     16
     """
+    line = line.rstrip('\n\r')
     if '\t' not in line:
         return len(line) - len(line.lstrip())
     result = 0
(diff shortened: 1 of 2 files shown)


pycodestyle.py
```diff
@@ -1794,6 +1794,7 @@ def expand_indent(line):
     >>> expand_indent('        \t')
     16
     """
+    line = line.rstrip('\n')
```
asottile: do we have to look out for `\r\n` (windows newlines)?

anntzer: I think probably not, python normalizes newlines at source read time?  OTOH this doesn't cost much and I doubt it can ever be wrong to strip `\r` so so let's do it too.

asottile: python2 does not (still supported here sadly) -- and python3 only does in text mode and if newline translation is enabled (it is by default)

PR #920, opened 2020-03-24
Instrument code coverage

asottile:
this makes it more clear which parts are covered / uncovered such that we can improve the testsuite to cover them

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,40 @@
+[run]
+source = .
+branch = true
+parallel = true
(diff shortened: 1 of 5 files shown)


asottile:
hmmm python3.4... right, I'll have to rework this a bit to avoid `covdefaults` which is a bit unfortunate

asottile:
travis ci seems to be having problems, so apologies as I do a bunch of noop force pushes to hopefully kick it :(

PR #921, opened 2020-03-27
Moved the metadata into setup.cfg

KOLANICH:


pyproject.toml
```diff
@@ -0,0 +1,6 @@
+[build-system]
+requires = ["setuptools>=42", "wheel", "read_version[toml] >= 0.3.0"]
+build-backend = "setuptools.build_meta"
+
+[tool.read_version]
+version = "pycodestyle:__version__"
```
asottile: I'm -1 on adding pyproject.toml at this time since it changes how users need to install this

you can get the same thing in `setup.cfg` by using `attr: pycodestyle.__version__`

setup.cfg
```diff
@@ -2,7 +2,51 @@
 universal = 1
 
 [metadata]
+name = pycodestyle
+author = Johann C. Rocholl
+author_email = johann@rocholl.net
+maintainer = Ian Lee
+maintainer_email = IanLee1521@gmail.com
+license = Expat
```
asottile: I know it's not a regression, but as far as I can tell the values for this field should be spdx values, of wich "MIT" would be the appropriate value here

setup.cfg
```diff
@@ -2,7 +2,51 @@
...
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: Implementation :: CPython
+    Programming Language :: Python :: Implementation :: PyPy
+    Topic :: Software Development :: Libraries :: Python Modules
+
+[options]
+py_modules = pycodestyle
+zip_safe = False
+install_requires = setuptools>=42; wheel; read_version[toml] >= 0.3.0
```
asottile: this is formatted incorrectly, also it's wrong -- pycodestyle has no runtime dependencies

pyproject.toml
KOLANICH: >since it changes how users need to install this

`python3 ./setup.py bdist_wheel` should work fine, as `pip install .` do.

>you can get the same thing in setup.cfg by using attr: pycodestyle.__version__

I don't think so. We should never import the package being built from its setup script. It would either break or interfere to building having as cwd a dir other than git repo root.

asottile: YAGNI also, you can do this directly with setup.cfg

PR #931, opened 2020-05-01
E714: fix chained `is not`

### LINKED ISSUE #767: E714 miscategorizes chained 'is not'
Python 3.6.5, pycodestyle 2.4.0

Misinterprets the first `is not`'s `not` as part of a `not foo is ...` expression.

test.txt:
```py
foo is not bar is not baz
```
output:
```
test:1:8: E714 test for object identity should be 'is not'
```
(not that chained `is not`s would ever appear in real/good code, but still worth noting)

suzil:
Closes #767

### DIFF
--- a/testsuite/E71.py
+++ b/testsuite/E71.py
@@ -64,6 +64,9 @@ if not X is Y:
 #: E714
 if not X.B is Y:
     pass
+#: E714
+if not X is Y is not Z:
+    pass
 
 #
 #: Okay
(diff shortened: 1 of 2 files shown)


testsuite/E71.py
```diff
@@ -64,6 +64,9 @@
 #: E714
 if not X.B is Y:
     pass
+#: E714
+if not X is Y is not Z:
```
asottile, line 68: it's a bit silly but does this still trigger E714 (I think it should, even though it's absurd)?

```python
if not X is not Y:
```

suzil: Good catch! I've adjusted the regex and added it as a test case.

So it will now catch the pattern `not X is/in` like before except when it's preceded by the keyword `is`.

PR #950, opened 2020-07-09
recognize cimport as a valid import statement

rongou:
`cimport` is used by Cython to import other modules. This PR adds support to recognize it so not to generate E402 errors when `cimport` is used.

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -1135,7 +1135,8 @@ def module_imports_on_top_of_file(
     if noqa:
         return
     line = logical_line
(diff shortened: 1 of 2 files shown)


asottile:
`pycodestyle` is a linter for python syntax, cython is not python syntax so I do not think this should be merged (you'd have similar problems for E999, cdef, etc. etc.)

sigmavirus24:
@asottile is correct

sigmavirus24:
It won't. Make a different linter for that

PR #954, opened 2020-07-21
Add lines arg to ast checks

cristiana-iordache:
When using AST for finding a specific Call, yielding the AST node's lineno and offset will mark the initial row if it's a multiline statement. If the multiline has many lines it becomes hard to pinpoint the exact place where the offending Call is. To solve this, it's necessary to add a physical parsing of the lines that form the multiline statement. For this, the lines of the source code need to be given as an argument.

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -2114,13 +2114,15 @@ class Checker(object):
         except (ValueError, SyntaxError, TypeError):
             return self.report_invalid_syntax()
         for name, cls, __ in self._ast_checks:
+            checker = None
             if len(_get_parameters(cls.__init__)) == 2:
                 checker = cls(tree, self.filename)
             elif len(_get_parameters(cls.__init__)) == 3:
                 checker = cls(tree, self.filename, self.lines)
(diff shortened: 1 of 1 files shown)


pycodestyle.py
```diff
@@ -2114,7 +2114,11 @@ def check_ast(self):
         except (ValueError, SyntaxError, TypeError):
             return self.report_invalid_syntax()
         for name, cls, __ in self._ast_checks:
-            checker = cls(tree, self.filename)
+            checker = None
+            if len(_get_parameters(cls.__init__)) == 3:
+                checker = cls(tree, self.filename)
+            elif len(_get_parameters(cls.__init__)) == 4:
+                checker = cls(tree, self.filename, self.lines)
```
asottile, line 2121: pycodestyle doesn't use this form, this also potentially leaves `checker` as none

sigmavirus24:
Also, we don't want to do this based on parameter length even if we wanted to do this here. Like @asottile mentioned, Flake8 does this correctly. I'm also 99% certain that `lines` here isn't the full lines of the file anyway and isn't what you want.

PR #957, opened 2020-08-08
E704: Ignore def with "..." as its body as stub / @typing.overload

terencehonles:
Handle a common pattern used by Python's `@typing.overload`

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -1205,7 +1205,8 @@ def compound_statements(logical_line):
                               "def")
                 break
             if STARTSWITH_DEF_REGEX.match(line):
-                yield 0, "E704 multiple statements on one line (def)"
+                if line[found + 1:].strip() != '...':  # stub / typing overload
+                    yield 0, "E704 multiple statements on one line (def)"
             elif STARTSWITH_INDENT_STATEMENT_REGEX.match(line):
                 yield found, "E701 multiple statements on one line (colon)"
         prev_found = found
--- a/testsuite/E30not.py
+++ b/testsuite/E30not.py
@@ -177,13 +177,11 @@ def foo():
     # for no E30x being emitted.
     def bar(): pass
     def baz(): pass
-#: E704:8:1 E704:10:1
+#: Okay
 from typing import overload
 from typing import Union
 
 
-# This emits the (ignored-by-default) E704, but here we're testing
-# for no E30x being emitted.
 @overload
 def f(x: int) -> int: ...
 @overload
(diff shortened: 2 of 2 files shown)


asottile:
is this needed? E704 is disabled by default

terencehonles:
Would it be better to be a configured setting? It seems to be enabled by default in flake8, but I don't know if that's because an ignore list removes all the defaults. It doesn't seem to be a particularly expensive change and it is pretty strict about what it allows.

asottile:
you probably want to use `extend-ignore` instead of `ignore` with flake8 fwiw

terencehonles:
So, I'm guessing you'd rather not have this change? For this rule I'm probably more or less OK going with the default that I didn't realize I was overriding since the line length and other rules will probably keep the code from getting too hairy.

However, I was also contemplating on touching how many blank lines are required between `@override`s since 2 blank lines between overrides for a function is **really** wordy (1 for a method is still not "great", but it's definitely better than 2). Before making that change, which is definitely more complex than this change, I'd want to know if I should even bother.

What I'd be wanting to change is to allow:

```python
from typing import List, override, TypeVar

T = TypeVar('T')
U = TypeVar('U')


@override
def function(
        ...,
        paramN: T) -> T:
    ...
@override
def function(
        ...,
        paramN: U) -> List[T]:
    ...
def function(*args):
    ...
```

Instead of requiring:

```python
from typing import List, override, TypeVar

T = TypeVar('T')
U = TypeVar('U')


@override
def function(
        ...,
        paramN: T) -> T:
    ...


@override
def function(
        ...,
        paramN: U) -> List[T]:
    ...


def function(*args):
    ...
```

The rationale for this change is that the further the definitions are apart from each other the less they are reading like they are related to each other. Since type definitions are both for the reader and the tool they should be as close to the function they are documenting. This change is related to that change because you may also want to allow a more terse `@override`:

```python
from typing import List, override, TypeVar

@override
def function(value: str) -> str: ...
@override
def function(value: bytes) -> bytes: ...
def function(value):
    ...
```

sigmavirus24:
Closing and re-opening to trigger a new test build

tobiasdiez:
It would be nice if this PR could be merged.

We have enabled E704 since in general that's a convenient choice (for us). However, in the context of overloaded methods its indeed overkill to consider `...` to be a separate statement.

Avasam:
You can even see it being used that way in:
https://peps.python.org/pep-0008/
https://peps.python.org/pep-0484/

And is the go-to format for type stubs

PR #961, opened 2020-09-11
fix skipping of physical checks when file does not end in newline

### LINKED ISSUE #960: W191: missed when file ends without a newline
I'm working on a project that enables flake8 only for testing for tabs for indentation (they _really_ don't want tabs).

In working on a minimal test to demonstrate the effectiveness of the check, I created a small Python file with tabs for indentation, then ran flake8 on it, only to find that it passed the check:

```
$ python -c "import pathlib; pathlib.Path('tabbed.py').write_text('def foo():\\n\\tpass')"
$ pip-run -q pycodestyle -- -m pycodestyle --select=W191 tabbed.py
$
```

But add a newline to the file and it fails the check:

```
$ python -c "import pathlib; pathlib.Path('tabbed.py').write_text('def foo():\\n\\tpass\\n')"
$ pip-run -q pycodestyle -- -m pycodestyle --select=W191 tabbed.py
tabbed.py:2:1: W191 indentation contains tabs
```

Both files contain tabs, but the former manages not to catch the error.

asottile:
Resolves #960

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -264,7 +264,7 @@ def trailing_blank_lines(physical_line, lines, line_number, total_lines):
     However the last line should end with a new line (warning W292).
     """
     if line_number == total_lines:
-        stripped_last_line = physical_line.rstrip()
+        stripped_last_line = physical_line.rstrip('\r\n')
         if physical_line and not stripped_last_line:
             return 0, "W391 blank line at end of file"
         if stripped_last_line == physical_line:
@@ -2125,21 +2125,30 @@ class Checker(object):
             self.report_error(1, 0, 'E902 %s' % self._io_error, readlines)
         tokengen = tokenize.generate_tokens(self.readline)
         try:
+            prev_physical = ''
             for token in tokengen:
                 if token[2][0] > self.total_lines:
                     return
                 self.noqa = token[4] and noqa(token[4])
-                self.maybe_check_physical(token)
+                self.maybe_check_physical(token, prev_physical)
                 yield token
+                prev_physical = token[4]
         except (SyntaxError, tokenize.TokenError):
             self.report_invalid_syntax()
 
-    def maybe_check_physical(self, token):
+    def maybe_check_physical(self, token, prev_physical):
         """If appropriate for token, check current physical line(s)."""
         # Called after every token, but act only on end of line.
+
+        # a newline token ends a single physical line.
         if _is_eol_token(token):
-            # Obviously, a newline token ends a single physical line.
-            self.check_physical(token[4])
+            # if the file does not end with a newline, the NEWLINE
+            # token is inserted by the parser, but it does not contain
+            # the previous physical line in `token[4]`
+            if token[4] == '':
+                self.check_physical(prev_physical)
+            else:
+                self.check_physical(token[4])
         elif token[0] == tokenize.STRING and '\n' in token[1]:
             # Less obviously, a string that contains newlines is a
             # multiline string, either triple-quoted or with internal
--- a/testsuite/W29.py
+++ b/testsuite/W29.py
@@ -9,6 +9,11 @@ class Foo(object):
 #: W291:2:35
 '''multiline
 string with trailing whitespace'''   
+#: W291 W292 noeol
+x = 1   
+#: W191 W292 noeol
+if False:
+	pass  # indented with tabs
 #: W292:1:36 noeol
 # This line doesn't have a linefeed
 #: W292:1:5 E225:1:2 noeol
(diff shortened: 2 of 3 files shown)


pycodestyle.py
```diff
@@ -2125,21 +2125,30 @@ def generate_tokens(self):
...
                 self.noqa = token[4] and noqa(token[4])
-                self.maybe_check_physical(token)
+                self.maybe_check_physical(token, prev_physical)
                 yield token
+                prev_physical = token[4]
         except (SyntaxError, tokenize.TokenError):
             self.report_invalid_syntax()
 
-    def maybe_check_physical(self, token):
+    def maybe_check_physical(self, token, prev_physical):
```
sigmavirus24, line 2139: We may need to replicate some of this logic to Flake8

jaraco:
Confirmed it works for the reported case (which is my actual use-case):

```
draft $ python -c "import pathlib; pathlib.Path('tabbed.py').write_text('def foo():\\n\\tpass')"
draft $ pip-run -q --use-pep517 git+https://github.com/asottile/pycodestyle@physical_checks_last_line -- -m pycodestyle tabbed.py
flake8 3.7.9 requires pycodestyle<2.6.0,>=2.5.0, but you'll have pycodestyle 2.6.0 which is incompatible.
tabbed.py:2:1: W191 indentation contains tabs
tabbed.py:2:6: W292 no newline at end of file
```

pycodestyle.py
```diff
@@ -2125,21 +2125,30 @@ def generate_tokens(self):
             self.report_error(1, 0, 'E902 %s' % self._io_error, readlines)
         tokengen = tokenize.generate_tokens(self.readline)
         try:
+            prev_physical = ''
```
jaraco, line 2128: Consider 2fc1d07, which re-uses a common 'windowing' pattern to limit the touch points of this change.

asottile: considered but it's wrong, for instance for blank lines it will run the previous physical line

jaraco: Oh, yuck. Do you mean the `if token[2][0] > self.total_lines`? Since the tests pass with that change, does that suggest the tests are missing an important case?

pycodestyle.py
asottile: indeed, I plan to work on that today!

pycodestyle.py
asottile: nah, for like:

```console
$ python3 -m tokenize /dev/stdin <<< $'x = 1\n\nx = 2\n'
0,0-0,0:            ENCODING       'utf-8'        
1,0-1,1:            NAME           'x'            
1,2-1,3:            OP             '='            
1,4-1,5:            NUMBER         '1'            
1,5-1,6:            NEWLINE        '\n'           
2,0-2,1:            NL             '\n'           
3,0-3,1:            NAME           'x'            
3,2-3,3:            OP             '='            
3,4-3,5:            NUMBER         '2'            
3,5-3,6:            NEWLINE        '\n'           
4,0-4,1:            NL             '\n'           
5,0-5,0:            ENDMARKER      ''   
```

the `NL` token would physical-check the line before it

asottile: (either way, I wouldn't consider some (imo very unreadable) itertools over a simple assignment)

jaraco: I'm afraid I don't understand where it goes wrong. Would you be willing to help me understand? From my understanding of the code, the only place this suggested change could go wrong was if `last[4]` raised a `TypeError` due to `last = None`. Is that where it goes wrong? It seems to me in the other cases, the behavior should be the same as with the `prev_physical` variable.

Is there a reason the test suite doesn't include a test that covers this expectation?

jaraco: Fair enough. Nevermind. I'm sure you know what you're doing.

pycodestyle.py
asottile: here's the flake8 patch with essentially the same logic!

https://gitlab.com/pycqa/flake8/-/merge_requests/451

pycodestyle.py
```diff
@@ -2125,21 +2125,30 @@ def generate_tokens(self):
...
         if _is_eol_token(token):
-            # Obviously, a newline token ends a single physical line.
-            self.check_physical(token[4])
+            # if the file does not end with a newline, the NEWLINE
+            # token is inserted by the parser, but it does not contain
+            # the previous physical line in `token[4]`
+            if token[4] == '':
+                self.check_physical(prev_physical)
+            else:
+                self.check_physical(token[4])
```
FichteFoll, line 2151: Nit
```suggestion
            self.check_physical(token[4] or prev_physical)
```

asottile: I disagree, the if statement is much more explicit that there's a special case here (and exactly what the special case is)

PR #962, opened 2020-09-26
introduce root argument

jdknight:
The following commit introduces a new argument `--root`, which allows a caller to specify the base location where a project-specific configuration can be found. This is primarily to enable callers which feed data through `stdin` to specify a possible project configuration without explicitly configuring the working directory when pycodestyle is invoked.

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -149,6 +149,8 @@ Quick help is available on the command line::
                          separated patterns (default: .svn,CVS,.bzr,.hg,.git)
     --filename=patterns  when parsing directories, only check filenames matching
                          these comma separated patterns (default: *.py)
+    --root=dir           when looking for project options, only check the
+                         directory provided (ignored if not set)
     --select=errors      select errors and warnings (e.g. E,W6)
     --ignore=errors      skip errors and warnings (e.g. E4,W)
     --show-source        show source code for each error
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -2556,6 +2556,9 @@ def get_parser(prog='pycodestyle', version=__version__):
                       help="when parsing directories, only check filenames "
                            "matching these comma separated patterns "
                            "(default: %default)")
+    parser.add_option('--root', metavar='dir',
+                      help="when looking for project options, only check the "
+                           "directory provided (ignored if not set)")
     parser.add_option('--select', metavar='errors', default='',
                       help="select errors and warnings (e.g. E,W6)")
     parser.add_option('--ignore', metavar='errors', default='',
@@ -2613,6 +2616,7 @@ def read_config(options, args, arglist, parser):
     config = RawConfigParser()
 
     cli_conf = options.config
+    root_dir = options.root
 
     local_dir = os.curdir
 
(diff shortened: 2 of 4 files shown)


asottile:
can you describe the usecase? I feel like this isn't necessary because you can change directories so my initial thought is -1 on adding an option that's not needed

jdknight:
The use case is related to the issue captured from an issue with PyCharm:

> How to disable specific PEP8 warnings in PyCharm at the repository level?
> https://stackoverflow.com/questions/63982795

I was attempting to disable a subset of PEP8 checks in a project's implementation through the PyCharm IDE. The IDE uses `pycodestyle` for PEP8 processing. It was observed that PyCharm would support user-specific configurations (e.g. `~\.pycodestyle`) but not project-specific configurations. Examining [how the IDE invokes pycodestyle](https://github.com/JetBrains/intellij-community/blob/webstorm/193.5233.13/python/src/com/jetbrains/python/validation/Pep8ExternalAnnotator.java#L204-L222) shows that it will feed a source's contents through `stdin` for information. While feeding into `stdin` without changing the working directory to the project's root directory, pycodestyle cannot find a repository's `setup.cfg` or `tox.ini` files (if any exist).

It is PyCharm's own issue for not properly supporting pycodestyle's project-specific configurations, as it could do this by attempting to change the working directory before spawning pycodestyle (at least, what was observed when crudely manipulating pycodestyle's implementation distributed inside PyCharm's installation folder). However, tools invoking pycodestyle may not have the existing flexibility of changing the working context of an invoked process. For example, browsing the API PyCharm uses to spawn command line calls, there does not appear to be a way to change the working directory of a spawned process.

While I can agree that this is not fault of pycodestyle, but it would be nice to provide callers (whether it be PyCharm or some other tool) the flexibility to define a path where project-specific configurations are expected to be held.

asottile:
doesn't `--config` (already existing option) solve this better? am I missing something

jdknight:
> doesn't `--config` (already existing option) solve this better? am I missing something

While `--config` provides the ability for a caller (e.g. tool) to load an explicit configuration file, the caller would also need to "know" the concepts of what `setup.cfg` or `tox.ini` are, check for their existence and then pass the file path into pycodestyle (if any).

By having an option such as `--root`, the caller does not need to know what project-specific configurations are supported by pycodestyle (whether it be `setup.cfg`, `tox.ini` or some other future configuration file). The tool can say to pycodestyle, "Hey, this is the root of a Python project that I want to work with, can you load any project-specific configuration options from there?".

asottile (changes requested):
yeah I'm more convinced this isn't necessary, pycharm should set the correct working dir or use `--config` -- pycodestyle shouldn't grow complexity for this

asottile:
actually, I'm not convinced this is even possible as you describe it, `pycodestyle` [already searches from the commonprefix of its arguments upwards to find the configuration](https://github.com/PyCQA/pycodestyle/blob/a1b1f7c795ace4c186fd76ec6df994deb716cea8/pycodestyle.py#L2613)

asottile:
yeah, here's a demo showing this issue is invalid

```console
$ cat setup.cfg 
[pycodestyle]
$ tree .
.
├── badfile.py
├── dir
│   └── goodfile.py
└── setup.cfg

1 directory, 3 files
$ (cd / && ~/opt/venv/bin/pycodestyle /tmp/y/badfile.py /tmp/y/dir/goodfile.py)
/tmp/y/badfile.py:1:2: E225 missing whitespace around operator
$ echo 'ignore = E225' >> setup.cfg 
$ (cd / && ~/opt/venv/bin/pycodestyle /tmp/y/badfile.py)
$
```

jdknight:
> yeah, here's a demo showing this issue is invalid

Using the above example but switching to `stdin` should show the issue:

```
$ (cd /tmp/y/ && cat /tmp/y/badfile.py | ~/opt/venv/bin/pycodestyle -)
$ (cd / && cat /tmp/y/badfile.py | ~/opt/venv/bin/pycodestyle -)
stdin:1:2: E225 missing whitespace around operator
```

I can agree that PyCharm _should_ be dealing with this by configuring the working directory to take advantage of pycodestyle's project-specific configurations. This pull request has only the intent in providing the flexibility for callers to not have to worry about the working directory when interacting with pycodestyle. That all being said, if this is a complexity which pycodestyle does not want to manage, that is fine by me.

sigmavirus24:
> That all being said, if this is a complexity which pycodestyle does not want to manage, that is fine by me.

Yeah, we provide ample ways for PyCharm to work around this. A `--root` parameter is likely only going to confuse existing users who genuinely will never need it.

I think the problem isn't _just_ complexity but also lack of necessity for literally any one else

PR #968, opened 2020-11-05
use github actions instead of travis-ci

asottile:


asottile:
#967 took close to **2 hours** to return results, let's see if github actions can do better (and test windows too!)

asottile:
oh hmmm, looks like actions might need to be enabled before I can try this 🤔

asottile:
@IanLee1521 or @sigmavirus24 can you enable github actions on this repository?

IanLee1521:
Unless this a race condition with @sigmavirus24 and they just beat me to it; it looks like Actions *are* enabled (below screenshot). Maybe Travis CI has to finish running first?

image

IanLee1521:
Hmm, I'm not sure that Travis is running... maybe because this is a WIP pull request?

sigmavirus24:
Also, I think actions need to be on master for them to run

asottile:
hmmm I've submitted PRs for actions before and they run immediately, even when I don't have commit bit -- I'll play around with this in my fork (it might be easier for me to control there?)

asottile:
had a few mistakes but it passes now on my fork: https://github.com/asottile/pycodestyle/runs/1359541910

sigmavirus24:
> hmmm I've submitted PRs for actions before and they run immediately, even when I don't have commit bit -- I'll play around with this in my fork (it might be easier for me to control there?)

That's never been my experience, frankly. I've always seen exactly this behaviour

.github/workflows/main.yml
```diff
@@ -0,0 +1,56 @@
+name: main
+on:
+  pull_request:
+  push:
+    branches: [master, 'test-me-*']
```
sigmavirus24, line 5: Probably don't want to keep `'test-me-*` in here

asottile: oh, I usually find this useful to debug CI without having to open a PR

asottile:
> 
> 
> > hmmm I've submitted PRs for actions before and they run immediately, even when I don't have commit bit -- I'll play around with this in my fork (it might be easier for me to control there?)
> 
> That's never been my experience, frankly. I've always seen exactly this behaviour

weird, a PR worked fine for me when I was adding a wheeling job here: https://github.com/hynek/argon2-cffi/pull/70

.github/workflows/main.yml
sigmavirus24: Stupid question: If you can push to a repo and you trust others who can push to the repo, why bother restricting this at all?

asottile: it's mostly so when pushing branches pushed to `pycqa/pycodestyle` and creating a pull request based on them you don't run the actions twice

sigmavirus24: :shrug: You're still running them twice, just once on a random push, and a second time when opening the PR. They're just separate.

asottile: in that case it would actually run 3 times:
- once when pushing
- once when opening the PR
- once when merged

the `branches: [...]` is to filter it down so it only runs twice:

- once when opening the PR
- once when merged

almost all of the CI systems get this wrong by default (imo) and spend twice as much time running tests than they need to be

sigmavirus24: :shrug: I don't personally see any problems with running tests that often. But also CI systems never actually test against a merged build, just against the head of the branch so they're usually trash either way

asottile: GA tests against the merged revision

sigmavirus24: It doesn't update when master moves forward in my testing though so it's a point-in-time merged revision, not one before you press merge

asottile: true true

sigmavirus24: Not an argument against GA just a general gripe

sigmavirus24 (approved):

Issue #953, opened 2020-07-16
E131 applies inconsistently with nested calls


mattdicarlo:
I often use an indentation format like this for long method chains with something like SQLAlchemy, which seems to be acceptable according to my reading of PEP8 and doesn't raise any errors when pycodestyle is run:

```python
rows = db.engine.execute(
    table.select()
         .where(table.c.column1 == 'something')
         .where(table.c.column2.is_(None))
         .order_by(table.c.column3)
         .limit(5)
)
```

However, if the name of the variable being assigned to is a particular length, `E131 - continuation line unaligned for hanging indent` is raised:

```python
aaaaaaaa = db.engine.execute(
    table.select()
         .where(table.c.column1 == 'something')  # E131
         .where(table.c.column2.is_(None))       # E131
         .order_by(table.c.column3)              # E131
         .limit(5)                               # E131
)
```

It seems to be related to having the `=` operator aligned with the indented lines? Adding just a single character to the left side of the assignment will remove the E131.

```python
aaaaaaaab = db.engine.execute(
    table.select()
         .where(table.c.column1 == 'something')
         .where(table.c.column2.is_(None))
         .order_by(table.c.column3)
         .limit(5)
)
```

Is this an issue in pycodestyle, or do I just not understand what's going on here?

PR #972, opened 2020-12-09
E131: indent_chances overwritten

andribas404:
Relates to https://github.com/PyCQA/pycodestyle/issues/953

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -638,6 +638,7 @@ def continued_indentation(logical_line, tokens, indent_level, hang_closing,
     hangs = [None]
     # visual indents
     indent_chances = {}
+    indent_chances_rows = {}
     last_indent = tokens[0][2]
     visual_indent = None
     last_token_multiline = False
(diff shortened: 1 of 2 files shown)


pycodestyle.py
```diff
@@ -706,6 +706,8 @@ def continued_indentation(logical_line, tokens, indent_level, hang_closing,
                 # ignore token lined up with matching one from a
                 # previous line
                 pass
+            elif isinstance(visual_indent, list) and text in visual_indent:
```
asottile, line 709: hmm so now this variable can be None, True, False, bytes, str, or a list of str?  this seems like we're trying to do too much with dynamicism here -- is there any way to make this easier to understand than overloading the type further?

pycodestyle.py
```diff
@@ -778,6 +780,13 @@ def continued_indentation(logical_line, tokens, indent_level, hang_closing,
             if start[1] not in indent_chances:
                 # allow lining up tokens
                 indent_chances[start[1]] = text
+            else:
+                v = indent_chances[start[1]]
+                if not isinstance(v, (bool, type(None))):
```
asottile, line 785: I'd rather see:

```
if list:
   ...
elif str / bytes:
    ...
```

are all of these branches covered by your tests?

PR #821, opened 2018-11-25
[WIP] Add https://pre-commit.com metadata

asottile:

PR #979, opened 2021-01-26
Re-consider providing pre-commit hook metadata

weibullguy:
I am opening this pull request for your re-consideration of PR #821.  I understand the reason for closing #821, but I don't use flake8 and I do use pycodestyle; perhaps I'm the only one in this situation (and need to re-consider using flake8).

My workflow is:

1. Run QA tools in my IDE on the file currently being edited.
2. Run QA tools across the group of files included in a commit using pre-commit hooks.
3. Run QA tools across the entire code base as a GitHub action on each pull-request/push.

Thus, step 2 requires a pycodestyle pre-commit hook.

Following the contributing guidelines, I've toxed my fork and all tests pass, although that's probably moot given the proposed change.  However, I've been using the following in my .pre-commit-config.yaml for the past three weeks with no problems.

```
- repo: https://github.com/weibullguy/pycodestyle
  rev: '2.6.1'
  hooks:
    - id: pycodestyle
      args: [--config, ./setup.cfg]
```

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,6 @@
+-   id: pycodestyle
+    name: pycodetyle
+    description: pycodestyle is a tool to check your Python code against some of the style conventions in PEP 8.
+    entry: pycodestyle
+    language: python
+    types: [python]
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,6 +1,6 @@
--   id: pycodestyle
-    name: pycodetyle
-    description: pycodestyle is a tool to check your Python code against some of the style conventions in PEP 8.
-    entry: pycodestyle
-    language: python
-    types: [python]
+- id: pycodestyle
+  name: pycodetyle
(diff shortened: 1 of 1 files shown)


asottile:
fwiw, I'm both the pre-commit maintainer and flake8 maintainer and I don't think this is helpful (and likely will confuse others into using pycodestyle when they really want to use flake8) -- that's why I closed my other PR adding it

unrelated, your `args: [--config, setup.cfg]` shouldn't be necessary as `pycodestyle` [should discover that configuration automatically](https://github.com/PyCQA/pycodestyle/blob/c47f5c1f977c20f16c19bc282834c70c9b40d800/pycodestyle.py#L96)

weibullguy:
That's fair and I get I'm probably outlying use case.  I'll close this PR too, but setup a cron job to keep my fork up to date and available with the .pre-commit-hook.yaml file in case anyone else out there wants to use it and stumbles across this PR.  Sorry for the noise.

BTW, I include this `args: [--config, setup.cfg]` out of an abundance of anal-retentiveness.

PR #987, opened 2021-04-01
detect multiple E712 in a line

### LINKED ISSUE #986: E712: False negative when multiple comparison in same line
Sample code:
```python
var = 1 if cond == True else -1 if cond == False else cond

if cond == False or cond == True:
    print(var)
```
When using pycodestyle 2.7.0
```
$ pycodestyle  mytest.py 
mytest.py:1:17: E712 comparison to True should be 'if cond is True:' or 'if cond:'
mytest.py:3:9: E712 comparison to False should be 'if cond is False:' or 'if not cond:'
```
When multiple equality comparisons are present in the same line, only first occurrence is detected.

asottile:
Resolves #986

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -1393,8 +1393,10 @@ def comparison_to_singleton(logical_line, noqa):
     None was set to some other value.  The other value might have a type
     (such as a container) that could be false in a boolean context!
     """
-    match = not noqa and COMPARE_SINGLETON_REGEX.search(logical_line)
-    if match:
+    if noqa:
+        return
+
+    for match in COMPARE_SINGLETON_REGEX.finditer(logical_line):
         singleton = match.group(1) or match.group(3)
         same = (match.group(2) == '==')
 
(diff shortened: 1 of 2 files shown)


friyaz (approved):

pycodestyle.py
```diff
@@ -1393,21 +1393,21 @@ def comparison_to_singleton(logical_line, noqa):
...
-        if singleton in ('None',):
-            code = 'E711'
-        else:
-            code = 'E712'
-            nonzero = ((singleton == 'True' and same) or
-                       (singleton == 'False' and not same))
-            msg += " or 'if %scond:'" % ('' if nonzero else 'not ')
-        yield match.start(2), ("%s comparison to %s should be %s" %
-                               (code, singleton, msg))
+    if not noqa:
```
sigmavirus24, line 1396: Can we change this logic a bit to make it easier to read?

```py
if noqa:
    return
for match in COMPARE_SINGLETON_REGEX.finditer(logical_line):
    # ...
```

sigmavirus24 (approved):
I have one non-urgent request that I won't block this on.

PR #989, opened 2021-05-02
Fix false-positive E211 with match and case

cdce8p:
Fix false-positive `E211 whitespace before '('` for `match` and `case`, Python 3.10

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -821,7 +821,10 @@ def whitespace_before_parameters(logical_line, tokens):
             # Allow "return (a.foo for a in range(5))"
             not keyword.iskeyword(prev_text) and
             # 'match' and 'case' are only soft keywords
-            prev_text not in ('match', 'case')
+            (
+                sys.version_info <= (3, 10) or
+                not keyword.issoftkeyword(prev_text)
+            )
         ):
             yield prev_end, "E211 whitespace before '%s'" % text
         prev_type = token_type
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -811,14 +811,18 @@ def whitespace_before_parameters(logical_line, tokens):
     prev_type, prev_text, __, prev_end, __ = tokens[0]
     for index in range(1, len(tokens)):
         token_type, text, start, end, __ = tokens[index]
-        if (token_type == tokenize.OP and
+        if (
+            token_type == tokenize.OP and
             text in '([' and
             start != prev_end and
             (prev_type == tokenize.NAME or prev_text in '}])') and
             # Syntax "class A (B):" is allowed, but avoid it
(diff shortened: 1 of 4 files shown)


pycodestyle.py
```diff
@@ -811,14 +811,18 @@ def whitespace_before_parameters(logical_line, tokens):
...
             start != prev_end and
             (prev_type == tokenize.NAME or prev_text in '}])') and
             # Syntax "class A (B):" is allowed, but avoid it
             (index < 2 or tokens[index - 2][1] != 'class') and
-                # Allow "return (a.foo for a in range(5))"
-                not keyword.iskeyword(prev_text)):
+            # Allow "return (a.foo for a in range(5))"
+            not keyword.iskeyword(prev_text) and
+            # 'match' and 'case' are only soft keywords
+            prev_text not in ('match', 'case')
```
asottile, line 824: perhaps `sys.version_info >= (3, 9) and keyword.issoftkeyword(prev_text)`

https://docs.python.org/3.10/library/keyword.html#keyword.issoftkeyword

testsuite/python310.py
```diff
@@ -0,0 +1,12 @@
+var, var2 = 1, 2
+#: Okay
+match (var, var2):
+    #: Okay
+    case [2, 3]:
+        pass
+    #: Okay
+    case (1, 2):
+        pass
+    #: Okay
```
asottile, line 10: the `#:` is supposed to split separate cases -- you shouldn't need anything except the first comment (also please include the `var1, var2` in the example)

cdce8p: 👌🏻 Didn't know that. Is there a way to run the test case only for Python 3.10+? Especially with the first change you suggested, the tests will fail without it.

asottile: the file naming is supposed to handle that, but it looks like this needs to have a `+` added to the second digit group here: https://github.com/PyCQA/pycodestyle/blob/c4460cb9dffbd9b7ee000dc31388d5f45fe484a0/testsuite/support.py#L172

asottile (approved):

cdce8p:
Should I add 3.10 to Github CI action as well?

asottile:
> 
> 
> Should I add 3.10 to Github CI action as well?

yeah that would be good! thanks

cdce8p:
> > Should I add 3.10 to Github CI action as well?
> 
> yeah that would be good! thanks

Done. All tests passed 🚀

east825:
Hi there! @asottile Do you plan to make a release with bug fixes for pattern matching support (this one in particular) before Python 3.10 comes out?

asottile:
please don't ask for releases, they will happen when they happen

these are not bugfixes, this is a feature and it is not complete (there are still issues with `match`)

PR #990, opened 2021-05-02
Add whitespace checks for ``match`` and ``case``

cdce8p:
`E271 multiple spaces after keyword` and `E275 missing whitespace after keyword` should also also be emitted for `match` and `case`. However, since they are only soft keywords, some additional logic is needed.

If accepted and the next version is released, this check could be added to flake8 as well ([flake8/setup.cfg](https://github.com/PyCQA/flake8/blob/7a371265cf876736b8d03907df26b813874b4201/setup.cfg#L72-L74)).

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -494,6 +494,17 @@ def whitespace_around_keywords(logical_line):
         elif len(after) > 1:
             yield match.start(2), "E271 multiple spaces after keyword"
 
+    if sys.version_info >= (3, 10):
+        match = MATCH_CASE_REGEX.match(logical_line)
+        if match:
+            whitespace = match.groups()[0]
+            if whitespace == ' ':
+                return
+            if whitespace == '':
+                yield match.start(1), "E275 missing whitespace after keyword"
+            else:
+                yield match.start(1), "E271 multiple spaces after keyword"
+
 
 @register_check
 def missing_whitespace_after_import_keyword(logical_line):
@@ -511,32 +522,6 @@ def missing_whitespace_after_import_keyword(logical_line):
         if -1 < found:
             pos = found + len(indicator) - 1
             yield pos, "E275 missing whitespace after keyword"
-
-
-@register_check
-def missing_whitespace_after_match_case(logical_line):
-    r"""Check whitespace after 'match' and 'case'.
-
-    Python 3.10
-    Okay: match status:
-    E271: match  status:
-    E271: case\tstatus:
-    E271: case   _:
-    E275: matchstatus:
-    E275: casestatus:
-    E275: case_:
-    """
-    if sys.version_info < (3, 10):
-        return
-    match = MATCH_CASE_REGEX.match(logical_line)
-    if match:
-        whitespace = match.groups()[0]
-        if whitespace == ' ':
-            return
-        if whitespace == '':
-            yield match.start(1), "E275 missing whitespace after keyword"
-        else:
-            yield match.start(1), "E271 multiple spaces after keyword"
 
 
 @register_check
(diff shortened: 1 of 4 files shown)


pycodestyle.py
```diff
@@ -512,6 +513,32 @@ def missing_whitespace_after_import_keyword(logical_line):
             yield pos, "E275 missing whitespace after keyword"
 
 
+@register_check
+def missing_whitespace_after_match_case(logical_line):
```
asottile, line 517: I would prefer if the existing checks were extended (this will provide better compatibility and performance)

testsuite/support.py
```diff
@@ -112,8 +113,14 @@ def selftest(options):
...
     checks = options.physical_checks + options.logical_checks
     for name, check, argument_names in checks:
+        python_version = None
         for line in check.__doc__.splitlines():
             line = line.lstrip()
+            match = SELFTEST_PY_REGEX.match(line)
+            if match:
+                python_version = tuple(map(int, match.groups()))
+            if python_version and sys.version_info < python_version:
+                continue
```
asottile, line 123: leftover / dead code ?

pycodestyle.py
cdce8p: That was my initial idea. However, at least AFAIK it isn't that simple. As example, the best place would probably have been the `whitespace_around_keywords` check. The issue there is that it also checks for whitespace before the keyword and that `match` and `case` are only soft keywords.

It might make sense to rename the new check to something like `whitespace_around_soft_keyword`. That would at least provide options to extend it in the future.

https://github.com/PyCQA/pycodestyle/blob/61138ca629242a998b495ce05b5cd25df2429900/pycodestyle.py#L473-L494

testsuite/support.py
cdce8p: No. The check requires `3.10`, at least the way it's currently designed. This is needed to only run the `doctest` for Python 3.10 and up.

cdce8p: https://github.com/PyCQA/pycodestyle/blob/11c41a490ad42132c31d0ec9a331111a3535ed6b/pycodestyle.py#L520-L527

asottile: I would either (1) remove the doctests and put them in actual tests files or (2) remove the specializating in the implementation

pycodestyle.py
asottile: I don't see why you can't add it to the bottom of the code you've linked?  renaming or adding functions leads to a bunch of annoying issues created by operating system distributors that intentionally and erroneously violate dependency constraints so I'd really like to avoid adding / renaming functions when possible

cdce8p: Ok, done.

pycodestyle.py
```diff
@@ -493,6 +494,17 @@ def whitespace_around_keywords(logical_line):
         elif len(after) > 1:
             yield match.start(2), "E271 multiple spaces after keyword"
 
+    if sys.version_info >= (3, 10):
+        match = MATCH_CASE_REGEX.match(logical_line)
+        if match:
+            whitespace = match.groups()[0]
```
asottile, line 500: `match[1]`

pycodestyle.py
```diff
@@ -493,6 +494,17 @@ def whitespace_around_keywords(logical_line):
         elif len(after) > 1:
             yield match.start(2), "E271 multiple spaces after keyword"
 
+    if sys.version_info >= (3, 10):
```
asottile, line 497: do things pass without the conditional?

cdce8p: Yeah, it does. The original idea was to not have to check the regex if it's not necessary.

asottile: 🤷 could go either way I guess

asottile (approved):

asottile:
this caused #1024 -- I'm just going to revert this for now -- we can potentially add it back later

PR #991, opened 2021-05-02
Small fix to version_info comparison

cdce8p:
Small issue I missed for #989. Technically this isn't a problem since even the alpha versions are `> (3, 10)`

/CC: @asottile

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -822,7 +822,7 @@ def whitespace_before_parameters(logical_line, tokens):
             not keyword.iskeyword(prev_text) and
             # 'match' and 'case' are only soft keywords
             (
-                sys.version_info < (3, 10) or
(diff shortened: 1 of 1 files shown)


pycodestyle.py
```diff
@@ -822,7 +822,7 @@ def whitespace_before_parameters(logical_line, tokens):
             not keyword.iskeyword(prev_text) and
             # 'match' and 'case' are only soft keywords
             (
-                sys.version_info <= (3, 10) or
+                sys.version_info < (3, 10) or
```
asottile, line 825: it should really be 3, 9 -- where the function was introduced (this will make cleanup maintenance earlier and easier in the future as well)

cdce8p: Didn't know that. Updated the PR

asottile (approved):

PR #992, opened 2021-05-04
improve performance of bare_except check

asottile:
- despite `re.compile(...)` being cached, this still amounts to 60% of the
  overhead of the `bare_except` check
- overall, this improved pycodestyle performance by about .25% (.0025)


sigmavirus24:
Am I reading your description correctly that this improves performance very little?

I'm not against it, I just want to make sure I am reading what you wrote correctly because it is a bit surprising.

sigmavirus24 (approved):

asottile:
> Am I reading your description correctly that this improves performance very little?
> 
> I'm not against it, I just want to make sure I am reading what you wrote correctly because it is a bit surprising.

yes, it improves the particular check by ~60% but makes very little difference overall (as there are many other checks)

Issue #945, opened 2020-06-12
E721: `Do not compare types, use 'isinstance()'` is overzealous with `type(None)`


ex-nerd:
Note: May be a dupe of (or at least related to) #882.

I discovered this while writing a decorator to parse the values of `typing.get_type_hints()`. This isn't exactly a bug in the flake8 feature, but rather an inconsistency in how `NoneType` is handled in the python3 typing system. This definitely feels like a bug, given that `isinstance(none_type, type(None)) != none_type is type(None)`. If nothing else it's worth documenting the behavior because I couldn't find anything written up about it:

Setup:
```python
# Setup:

from typing import Union

my_type = Union[str, None]
# --> typing.Union[str, NoneType]
none_arg = my_type.__args__[1]
# --> <class 'NoneType'>
```

Examples/behavior:

```python
# Expected output based on recommendations for validating NoneType in python 3,
# but this triggers e721:
none_arg is type(None)
# --> True

# Just for example:
none_arg is None
# --> False

# Recommended by E721 but not actually valid
isinstance(none_arg, type(None))
# --> False

isinstance(none_arg, None)
# --> TypeError: isinstance() arg 2 must be a type or tuple of types

isinstance(none_arg, (None, ))
# --> TypeError: isinstance() arg 2 must be a type or tuple of types
```

Current workaround:

```python
# Redefine local copy of NoneType no longer present in Python3
NoneType = type(None)

none_arg is NoneType
# --> True

# However, this fails to pass `mypy` inspection, so I'm now back to:
none_arg is type(None)  # noqa: E721
# --> True
```

FichteFoll:
```py
isinstance(None, none_arg)
```
This should yield true, but is unexpected in the grand scheme of things. `type(None)` should be equivalent to `str` and since `x is str` is allowed, so should `x is type(None)`.

ex-nerd:
@FichteFoll Interesting ... I hadn't thought about reversing the two properties in `isinstance()`as a workaround.

asottile:
by the way, [`NoneType` is back](https://docs.python.org/3.10/library/types.html#types.NoneType)

Delwin9999:
As a side note the reversed isinstance(None, X) doesn't work if X is an optional type (which is a Union[X, NoneType]).  The whole issubclass and isinstance breaking all over the place when it comes to optional types is irksome to work around.

asottile:
`x is type(None)` is allowed via the latest pycodestyle

PR #998, opened 2021-05-26
Allow explicit comparisons to type(None).

anntzer:
Closes https://github.com/PyCQA/pycodestyle/issues/945 (see in particular https://github.com/PyCQA/pycodestyle/issues/945#issuecomment-643750340).


anntzer:
Kindly bumping.

asottile:
I think this is less necessary due to the reintroduction of `types.NoneType` or the ability to assign

anntzer:
Kindly bumping again.  (I'd prefer have a decision in either direction than letting the PR linger.  I could also split out `type(None)` checks into a separate category that can be switched on and off independently of the others, if you prefer that.)

anntzer:
@asottile If you don't like this PR, can you clarify if you intend to close #945 as invalid?  (I wouldn't *like* that, but I accept that it's not my call; OTOH leaving it open gives an impression that it's something that should be addressed.)

anntzer:
Closing as inactive.  Feel free to reuse the patch if desired.

PR #1003, opened 2021-07-09
Fix false-positive with star pattern

### LINKED ISSUE #1002: False positive E225 on initial PEP 634 star patterns
It's reported only when a star pattern is a very first pattern in a case clause.

```
match [42]:
    case *_, 42:  # E255 missing whitespace around operator
        pass
```

cdce8p:
Similar to #989: add ignore for soft keywords.

Closes #1002

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -964,8 +964,13 @@ def missing_whitespace_around_operator(logical_line, tokens):
                 # Check if the operator is used as a binary operator
                 # Allow unary operators: -123, -x, +1.
                 # Allow argument unpacking: foo(*args, **kwargs).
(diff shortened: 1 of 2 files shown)


asottile (approved):
whew, thanks for improving that condition as well -- it was *really* difficult to understand before this patch

cdce8p:
Just noticed that the PR hasn't been merged yet. Do you need me to add anything else?

cdce8p:
Any update now that `3.10` has been released?

asottile:
we have a dual sign off process for merging PRs, but whatever ask for forgiveness rather than permission

PR #1006, opened 2021-07-25
pycodestyle --statistics for E501 reports the longest line

cclauss:
When I run pycodestyle on a new codebase, I want to know what I should set `--max-line-length` to be so that new code changes do not make the line length problem worse.  I currently need to scan the verbose output and manually search the E501 messages for the longest line and then run `pycodestyle --max-line-length=97` to see if I found the maximum value.  If I did not (often happens), then I need to rinse and repeat.  This is both error-prone and frustrating so, what if the E501 message of `pycodestyle --statistics` _always_ showed me the length of the longest line?  This PR makes that happen!

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -174,6 +174,7 @@ MATCH_CASE_REGEX = re.compile(r'^\s*\b(?:match|case)(\s*)(?=.*\:)')
 BLANK_EXCEPT_REGEX = re.compile(r"except\s*:")
 
 _checks = {'physical_line': {}, 'logical_line': {}, 'tree': {}}
+_longest_line = 0
 
 
 def _get_parameters(function):
@@ -316,6 +317,9 @@ def maximum_line_length(physical_line, max_line_length, multiline,
             except UnicodeError:
                 pass
         if length > max_line_length:
+            global _longest_line
+            if length > _longest_line:
+                _longest_line = length
             return (max_line_length, "E501 line too long "
                     "(%d > %d characters)" % (length, max_line_length))
 
@@ -2267,6 +2271,7 @@ class BaseReport(object):
     def __init__(self, options):
         self._benchmark_keys = options.benchmark_keys
         self._ignore_code = options.ignore_code
+        self._max_line_length = options.max_line_length
         # Results
         self.elapsed = 0
         self.total_errors = 0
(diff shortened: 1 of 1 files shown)


asottile:
I'm -1:

- the use of global is a nonstarter
- this isn't really the point of `--statistics`
- you can already get the same information pretty easily:

```console
$ git ls-files -- '*.py' | xargs flake8 --max-line-length=50 | grep E501 | cut -d: -f4 | sort -ru | head -1
 E501 line too long (79 > 50 characters)
```

sigmavirus24:
I strongly agree with @asottile. This isn't `pycodestyle --generate-baseline` this is `pycodestyle --statistics`. The former would tell you "This is the longest line" (in theory) and the latter tells you how frequently certain errors are seen. There's a large difference. You could, manually, interpret statistics output to create your baseline configuration, but that's not its purpose first and foremost.

cclauss:
The goal we really to effect the output of `flake8` but that does not have a `--generate-baseline` option.  Other thoughts?

sigmavirus24:
None of the tools have a `--generate-baseline` option (aside from something similarly named in `bandit`).

Also, Flake8 does its own statistics because pycodestyle is but one set of plugins for it. So to influence Flake8's output, this is not going to work. Further, if Flake8 did lean on pycodestyle for this, multiprocessing would render it inconsistent, incorrect, and useless.

I've often thought of building a tool to suggest/work-through (terminal dialogs) the JSON output of Flake8 and then print out a suggested configuration based on what the user wants. That doesn't exist, however. You could build that.

cclauss:
Agreed that `bandit` makes it easy...
```python
with open("bandit.out.txt") as in_file:
    errors = sorted(set(line[11:15] for line in in_file if line.startswith(">> ")))
    print(f"bandit --recursive --skip {','.join(errors)} .")
```

sigmavirus24:
I'm fairly certain bandit had a command/option to build their config for you and that you don't need to generate a command to be run. But I can't tell if you're just messing with me now.

Like I said, flake8 has plenty of ways to format the output and those can be used to generate a config file.

cclauss:
Sorry...  Not messing around with you.  I did work out something with flake8...  I will keep testing that.

Issue #813, opened 2018-10-29
Support pyproject.toml for configuration


ZeeD26:
I would love to see support for `pyproject.toml` for the configuration as, e.g., [black does](https://github.com/ambv/black#configuration-format)

IanLee1521:
@ZeeD26 - would you be willing to submit a PR?

sigmavirus24:
@IanLee1521 the configuration management in this project is an abysmal nightmare. Until someone has time to do serious refactoring, I strongly advise against accepting this as a feature.

IanLee1521:
Good point.

asears:
Is there more traction now for pyproject.toml format?  Seems more prevalent on Github.  We are looking at Poetry and this would be useful to have central configuration.

sigmavirus24:
Not really, no

carstencodes:
Hi, I must agree to @sigmavirus24, the configuration is really messed up, but I think I found a simple and backward compliant way of adding this feature.

The idea is to read the configuration from the pyproject.toml into a dictionary and add it to the configparser object.

To keep the current behavior, that no dependencies must be used, the feature is added as an extra.

There are still some things to do before #1007 can be merged:

* Fallback mode: Read data from setup.cfg/tox.ini if no data is found in pyproject.toml
* Add Unit tests for pyproject.toml

Please tell me, what you think.

PR #1007, opened 2021-08-04
WiP: Added PEP518 support

carstencodes:
The PIP518 support as described in
https://github.com/PyCQA/pycodestyle/issues/813
has been implemented roughly.

It uses the toml package, since the toml package describes itself as
being Python 2 compatible.
Using another package will create  a dependency that
might not be wanted. Hence the dependency is
treated optional as an extra, so consumers can decide
which revision they would like to use.

TODO: Unit test must be written explicitely.


OscarCreator:
Any update when this will be fixed?

PR #1016, opened 2021-09-16
Issue #582: disallow tabs before inline comments

### LINKED ISSUE #582: E261 ignores tab/space mix
```
printf 'foo = "bar" \t# comment\n' | pep8 -
```
I expected the following output but got nothing:
```
stdin:1:12: E261 at least two spaces before inline comment
```

spaceone:
Detect invalid tabs before inline comments:
```
printf 'foo = "bar" \t# comment\n' | pycodestyle -
stdin:1:12: E261 at least two spaces before inline comment
printf 'foo = "bar"\t # comment\n' | pycodestyle -
stdin:1:12: E261 at least two spaces before inline comment
printf 'foo = "bar"\t\t# comment\n' | pycodestyle -
stdin:1:12: E261 at least two spaces before inline comment
```

Fixes #582


spaceone:
any reviewers?

asottile:
I don't think this is correct, a space + a tab is "at least two spaces"

Issue #1018, opened 2021-09-27
E401: Module Level Dunder Names with Type Hints


randolf-scholz:
According to https://www.python.org/dev/peps/pep-0008/#module-level-dunder-names:

> Module level "dunders" (i.e. names with two leading and two trailing underscores) such as __all__, __author__, __version__, etc. should be placed after the module docstring but before any import statements except from __future__ imports. 

I noticed that `pycodestyle` violates this in several regards. Consider the following example:

```python
r"""My favourite circle constant."""

from __future__ import annotations

__all__ = ["tau"]

import math

tau: float = 2*math.pi
```
This passes fine. However, the following modifications do not:

1. If one puts the type hint `__all__: list[str] = ["tau"]`, then `E402 module level import not at top of file
` gets raises 
2. If one moves `__all__`'s definition to any other place in the script, no issue gets raised.

**Expected behaviour:**

1. Adding a type hint to `__all__` should not raise an error.
2. An error/warning should be raised if `__all__` is not in-between `__future__` and other import statements as per the corresponding PEP8 rule.

randolf-scholz:
Related to [https://github.com/python/mypy/issues/10103](Mypy complains about empty __all__ being untyped #10103). Either the convention should be that `__all__` is always untyped, but then `mypy` shouldn't raise an error if it is empty. In the current situation a very simple template like

```python
r"""Module summary Line.

Module description.
"""

__all__ = []

import logging

logger = logging.getLogger(__name__)
```

will make either `pycodestyle` complain with `E402` or `mypy`  with `[var-annotated]`, depending on whether a type hint is added to `__all__` or not.

PR #1019, opened 2021-09-30
Allow __all__ to be typed

sigmavirus24:
Closes gh-1018

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -169,7 +169,7 @@ STARTSWITH_INDENT_STATEMENT_REGEX = re.compile(
         'while',
     )))
 )
-DUNDER_REGEX = re.compile(r'^__([^\s]+)__ = ')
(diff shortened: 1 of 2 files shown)


pycodestyle.py
```diff
@@ -169,7 +169,7 @@ def lru_cache(maxsize=128):  # noqa as it's a fake implementation.
         'while',
     )))
 )
-DUNDER_REGEX = re.compile(r'^__([^\s]+)__ = ')
+DUNDER_REGEX = re.compile(r"^__([^\s]+)__(:\W+[a-zA-Z.0-9_\[\]\"]+)? = ")
```
asottile, line 172: should the `\W+` here be `\s*` ?

sigmavirus24: Probably, I was trying to fit this in in 15m after having just woken up so the caffeine hadn't hit yet

sigmavirus24: Updated and made it a non-capturing regex as well

asottile (approved):

PR #1032, opened 2021-10-13
Emit E221-E224 for walrus op

cdce8p:
Emit multiple spaces warning before / after assignment expression operator.

```py
#: E221:1:5 E222:1:9 E221:3:6
if x  :=  1:
    pass
if (x  := 2):
    pass
#: E223:1:5 E224:1:8
if x	:=	2:
    pass
```


cdce8p:
The force-push was to add another test case. I didn't expect the PR to be reviewed so soon. Sorry about that.

PR #1033, opened 2021-10-13
Add whitespace checks for `match` and `case` - 2nd attempt

cdce8p:
Second attempt after #990

I've improved the default `MATCH_CASE_REGEX`, especially the lookahead part.
```diff
- (?=.*\:)
+ (?=.*\S.*\:(?=\s*$|\s*#))
```
It now requires at least one non-whitespace char to be present between `case` and `:`. Furthermore it matches end of line or begin of comment.

To account for case statements written inline, I've added fallback logic. That uses the tokenizer to check the `:` isn't inside any brackets.
```py
match 0:
    case _: print('hi')
    case {"x": 42}: pass
    case "Hello: World": pass
```


cdce8p:
Closing this as there doesn't seem to be interest to implement it. Please feel free to reuse the code as you see fit.

PR #1035, opened 2021-11-06
Properly warn for E262 with non breaking whitespaces

### LINKED ISSUE #1034: False negative E262 inline comment should start with '# ' when there is a non breaking whitespace after a space
It happens often that you start typing a ``#`` then keep typing on the key longer than required and create a non breaking whitespace instead of a space. Then old version of ``black`` added a space at the beginning and there would be one space and one non-breaking white space (NBSP) at the start of the comment.

```python
a = 42 #  (One space one NBSP)
```
Should raise a E262 like
```python
a = 42 #  (Two spaces)
```
does.

I'm willing to contribute the evolution if this is a desirable change.

Pierre-Sassoulas:
Closes #1034


pycodestyle.py
```diff
@@ -131,7 +132,8 @@ def lru_cache(maxsize=128):  # noqa as it's a fake implementation.
     'and', 'in', 'is', 'or'] +
     FUNCTION_RETURN_ANNOTATION_OP +
     ASSIGNMENT_EXPRESSION_OP)
-WHITESPACE = frozenset(' \t')
+NON_BREAKING_WHITESPACE = ' '
```
asottile, line 135: please use the escape sequence instead otherwise this code is very difficult to read

Pierre-Sassoulas: Done in https://github.com/PyCQA/pycodestyle/pull/1035/commits/6975080d6b6d7cb647b35f8abdcbea6c986ba0c0

pycodestyle.py
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+# -*- coding: utf8 -*-
```
asottile, line 2: this shouldn't be needed now

pycodestyle.py
```diff
@@ -131,7 +132,8 @@ def lru_cache(maxsize=128):  # noqa as it's a fake implementation.
     'and', 'in', 'is', 'or'] +
     FUNCTION_RETURN_ANNOTATION_OP +
     ASSIGNMENT_EXPRESSION_OP)
-WHITESPACE = frozenset(' \t')
+NON_BREAKING_WHITESPACE = '\xa0'
```
asottile, line 135: this constant isn't used anywhere but below -- inline it

pycodestyle.py
Pierre-Sassoulas: https://github.com/PyCQA/pycodestyle/pull/1035/commits/61a8e66d5289f54155a4bc14dd1f0a7a9661315c

pycodestyle.py
Pierre-Sassoulas: https://github.com/PyCQA/pycodestyle/pull/1035/commits/41545a21e9f29b5c9ab46d45b89e3ed099647886

pycodestyle.py
Pierre-Sassoulas: I reverted the commit as we need the encoding for examples in docstrings.

asottile: the examples in the docstrings aren't readable, I'd prefer writing `\xa0` there as well

Pierre-Sassoulas: Would you rather have no docstring exemples with NBSP or some hard to read one ? Because ``    E262: x = x + 1  # \xa0Increment x`` does not trigger E262 (a real one will).

asottile: there's other examples where `\t` is used similarly, putting `E262: x = x + 1 # \xa0Increment x` (literally) into the docstring I believe gets the point across and is readable

Pierre-Sassoulas: There's a self test that check that the example are working apparently:
```
____________________________________________________________________________________ PycodestyleTestCase.test_selftest ____________________________________________________________________________________

self = <testsuite.test_all.PycodestyleTestCase testMethod=test_selftest>

    def test_selftest(self):
        fail_s, done_s = selftest(self._style.options)
        self.assertTrue(done_s, msg='tests not found')
>       self.assertFalse(fail_s, msg='%s failure(s)' % fail_s)
E       AssertionError: 1 is not false : 1 failure(s)

testsuite/test_all.py:33: AssertionError
------------------------------------------------------------------------------------------ Captured stdout call -------------------------------------------------------------------------------------------
pycodestyle.py: failed to find E262:
x = x + 1  # \xa0Increment x
```

asottile: if you remove the `r` on that docstring it should fix it

Pierre-Sassoulas: Done in https://github.com/PyCQA/pycodestyle/pull/1035/commits/fe23aa2344d6f6b81e968b60fb4597d3c0d61347

Pierre-Sassoulas:
@asottile any idea why the 2.7 tests fails when using ``\xa0`` ? They work if we do this https://github.com/PyCQA/pycodestyle/pull/1035/commits/693934309a317a98805e9c1db46bc63daf7ed717

asottile:
> @asottile any idea why the 2.7 tests fails when using `\xa0` ? They work if we do this [6939343](https://github.com/PyCQA/pycodestyle/commit/693934309a317a98805e9c1db46bc63daf7ed717)

good question, I vote we just drop python 2.7 -- I'll put together a branch if there isn't one already

sigmavirus24:
If you rebase @Pierre-Sassoulas your tests should pass now that we're not testing against 2.7

Pierre-Sassoulas:
This is rebased/cleaned up ! Thank you for the review, and thank you for dropping python 2.7, not having to deal with that encoding issue is a relief 😄

asottile (approved):

Issue #47, opened 2012-01-02
Deny direct variable type comparison


zyluo:
According to PEP8,

<pre>
- Object type comparisons should always use isinstance() instead
      of comparing types directly.

        Yes: if isinstance(obj, int):

        No: if type(obj) is type(1):

      When checking if an object is a string, keep in mind that it might be a
      unicode string too! In Python 2.3, str and unicode have a common base
      class, basestring, so you can do:

        if isinstance(obj, basestring):
</pre>

People, including me, tend to import the types module and do something like "type(XXX) is types.IntType".

Comparing types directly should trigger an error.

PR #1041, opened 2021-12-12
fix E721 types regex

asfaltboy:
tl;dr [The PEP8 example for the issue][0] only includes the `type(a) == type(b)` comparison, it does not include the `== types.SomeType` issue, the latter should not be handled under this error check.

This check was introduced to trunk following issue #47, and I believe it _incorrectly_ borrowed the example from the issue literally. The E721 check was since reported on a few false-positive issues and then expanded a number of times to handle those around types (see 034b9f4fb082f72e1087a8a27dd0dd61486d1888 and 0327e55e756e20ade468e67992f9f43da14aabb4).

I believe the behaviour should change because object comparison cannot be confirmed a "type comparison" in the same sense; not without having additional information, which is not possible with regex alone (_maybe_ with AST analysis).
This check also breaks many ligimitate cases (such as having your own types module with an Enum class for instance, and comparing a value to the  enum member).
If the community wants to keep it, I strongly advise it be made a separate warning level advisory (W###) rather than overloading E721, so we can handle any logic around it separately.

For reference, I learned more about the pattern, and tested for the cases using regexr:
* Original pattern - https://regexr.com/6bc9t
* Fixed pattern - https://regexr.com/6bca0

fixes issue #830

[0]: https://www.python.org/dev/peps/pep-0008/#programming-recommendations

### DIFF
--- a/testsuite/E72.py
+++ b/testsuite/E72.py
@@ -80,3 +80,8 @@ try:
     pass
 except Exception:
     pass
+#: Okay
+from . import custom_types as types
+
+red = types.ColorTypeRED
+red is types.ColorType.RED
--- a/testsuite/E72.py
+++ b/testsuite/E72.py
@@ -4,7 +4,7 @@ if type(res) == type(42):
 #: E721
 if type(res) != type(""):
     pass
-#: E721
+#: Okay
 import types
 
 if res == types.IntType:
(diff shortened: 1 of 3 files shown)


testsuite/E72.py
```diff
@@ -80,3 +78,8 @@ def func_histype(a, b, c):
     pass
 except Exception:
     pass
+#: Okay
+from . import custom_types as types
+
```
sigmavirus24, line 83: These test files aren't ever executed so we don't need a real module backing them here

asfaltboy: thanks, removed `custom_types.py` in 5df259b

sigmavirus24 (approved):

asottile:
reverting this for now so I can get a release out -- this caused significant churn -- I have a proposal in #1084 to make this more palatable when reintroducing it

Issue #1017, opened 2021-09-21
[feature] Add support for pre-commit


Freed-Wu:
<https://pre-commit.com/hooks.html>

Thanks!

asottile:
#821

PR #1044, opened 2021-12-22
Add pre-commit-hooks.yaml

### LINKED ISSUE #1043: Add pre-commit-hooks.yaml
Add pre-commit-hooks.yaml to enable usage of this utility in pre-commit checks

martinrakovec:
Resolves #1043

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,6 @@
+-   id: pycodestyle
+    name: pycodestyle
+    description: 'Python style guide checker'
+    entry: pycodestyle
+    language: python
+    types: [python]


.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,7 @@
+-   id: pycodestyle
+    name: pycodestyle
+    description: 'pycodestyle (formerly called pep8) - Python style guide checker'
+    entry: pycodestyle
+    language: python
+    types: [python]
+    require_serial: true
```
sigmavirus24, line 7: I don't believe `require_serial: true` makes any sense for pycodestyle.

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,7 @@
+-   id: pycodestyle
+    name: pycodestyle
+    description: 'pycodestyle (formerly called pep8) - Python style guide checker'
```
sigmavirus24, line 3: This should just be `pycodestyle - Python style guide checker` or `Python style guide checker`

.pre-commit-hooks.yaml
martinrakovec: You are right. I removed it

.pre-commit-hooks.yaml
martinrakovec: Thanks for your comment. I changed it.

asottile:
this was previously discussed in #821 #979 #1017

martinrakovec:
> 

@asottile Does it mean I should use flake8 instead of pycodestyle in my project and close this PR?

asottile:
yep!

PR #1056, opened 2022-02-24
docs: Fix double-dash converted to en-dash

wjandrea:
The double-dash for command line option `--ignore` was being converted to an en-dash.

I'm not very familiar with RST, but this change seems correct. By all means reject if not.

Same change as PyCQA/pydocstyle/pull/579


asottile (approved):
can confirm it's broken on https://pycodestyle.pycqa.org/en/latest/intro.html -- thanks!

PR #1067, opened 2022-05-07
Add SARIF support

shaopeng-gh:
Add support for output in SARIF format.

About SARIF format:
https://sarifweb.azurewebsites.net/

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -107,13 +107,16 @@ Or you can display how often each error was found::
   1188    W602 deprecated form of raising exception
 
 You can also make ``pycodestyle.py`` show the error text in different formats by
-using ``--format`` having options default/pylint/custom::
+using ``--format`` having options default/pylint/sarif/custom::
 
   $ pycodestyle testsuite/E40.py --format=default
   testsuite/E40.py:2:10: E401 multiple imports on one line
 
   $ pycodestyle testsuite/E40.py --format=pylint
   testsuite/E40.py:2: [E401] multiple imports on one line
+
+  $ pycodestyle testsuite/E40.py --format=sarif
+  return results in Static Analysis Results Interchange Format (SARIF)
 
   $ pycodestyle testsuite/E40.py --format='%(path)s|%(row)d|%(col)d| %(code)s %(text)s'
   testsuite/E40.py|2|10| E401 multiple imports on one line
@@ -162,7 +165,7 @@ Quick help is available on the command line::
     --indent-size=n      set how many spaces make up an indent (default: 4)
     --hang-closing       hang closing bracket instead of matching indentation of
                          opening bracket's line
-    --format=format      set the error format [default|pylint|<custom>]
+    --format=format      set the error format [default|pylint|sarif|<custom>]
     --diff               report only lines changed according to the unified diff
                          received on STDIN
 
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -48,6 +48,7 @@ W warnings
 """
 import bisect
 import inspect
+import json
 import keyword
 import os
 import re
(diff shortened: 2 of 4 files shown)


asottile:
this would probably be better done as a flake8 plugin rather than pycodestyle (which would help more broadly)

sigmavirus24:
I agree with @asottile

EasyRhinoMSFT:
Hello @asottile @sigmavirus24 if we add SARIF output support to flake8, would that capability then be available to pycodestyle? IOW, does pycodesyle make use of flake8?

asottile:
flake8 runs pycodestyle, yes

sigmavirus24:
> Hello @asottile @sigmavirus24 if we add SARIF output support to flake8, would that capability then be available to pycodestyle? IOW, does pycodesyle make use of flake8?

Flake8 runs pycodestyle, not the other way around

EasyRhinoMSFT:
I see, thanks for clarifying. I'll create an issue in the flake8 repo to get us started.

asottile:
that's the beauty of flake8, you don't have to! you can write your own formatter plugin

sigmavirus24:
@EasyRhinoMSFT an example is [PyCQA/flake8-json](/PyCQA/flake8-json) for formatting things via plugin. It might be a good place to fork from too for a flake8-sarif plugin that y'all can maintain and publish independently

EasyRhinoMSFT:
Thank you both again for your quick replies. Would either of these arrangements be possible?
Add our format to the flake8-json formatter (SARIF is a JSON schema)
or
Create a new repo under PyCQA

We think it makes the most sense to have our formatter as close to flake8 as possible.

Once our formatter is complete, you can publish GitHub workflows for your linters that are configured to use the SARIF formatter. We would be happy to create the workflow YML for you.
Publishing the workflow in your repos enables you to easily update the linter version when you do releases. You can also add entries to the [workflows page](https://github.com/topics/github-workflows), then add the topic to each repo to advertise.

sigmavirus24:
> Thank you both again for your quick replies. Would either of these arrangements be possible?
> Add our format to the flake8-json formatter (SARIF is a JSON schema)

I don't think it's a fit there.

> or
> Create a new repo under PyCQA

That's possible, but you would be the person maintaining it, not Anthony or me.

> We think it makes the most sense to have our formatter as close to flake8 as possible.

Most flake8 plugins are nowhere near it. It's unnecessary to keep things close.

> Once our formatter is complete, you can publish GitHub workflows for your linters that are configured to use the SARIF formatter. We would be happy to create the workflow YML for you.

Why would we bother? We don't need that output

> Publishing the workflow in your repos enables you to easily update the linter version when you do releases. You can also add entries to the [workflows page](https://github.com/topics/github-workflows), then add the topic to each repo to advertise.

What's the point of triggering new releases when flake8 releases when no other plugin needs to do this. We're not maintaining this for you, it probably doesn't need updates and we have no use for it ourselves.

EasyRhinoMSFT:
We would appreciate the creation of a repo called flake8-sarif under PyCQA.

Regarding workflows, the benefit is to increase usage of your linters by making them available for build-time static analysis scanning to every repo on GitHub. SA tools are required to output results in the SARIF format. One of our goals with this effort is to increase the usage of SARIF, so adding a formatter to flake8 and creating the workflows is mutually beneficial.

When I mentioned releases, I meant releases of flake8. When that occurs, workflows that use it should have their version dependency updated accordingly. This would be most reliably done on your side because you are the ones creating the release. If we are responsible, we may not be aware of a release, which could be disadvantageous in the case of security vulnerabilities for example.

shaopeng-gh:
in addition to @EasyRhinoMSFT's question above, after we settle the repo location, this is the code we are looking to use in the target repo currently: https://github.com/shaopeng-gh/flake8-sarif/pull/1
let me know if we are heading at the right direction, thanks!

PR #1078, opened 2022-06-22
Fixing false negatives with E731 lambda assignments

### LINKED ISSUE #1061: E731 not raised when assigned via tuple unpacking
```python
a = lambda x: x  # Raises E731
a, b = lambda x: x, lambda y: y  # Doesn't raise E731
a, b = 1, lambda y: y  # Doesn't raise E731
```

Sam-Cleveland:
fixes #1061 
Changed out the current implementation of lambda assignment detection for a regex based one. As stated in #1061, there are several lambda assignment cases that are being missed. Additionally, the current detection implementation is not in the most logical place. Before, detection of lambdas was tied to searching for ':' characters. Now lambda assignment statements are all just detected by:
```regex
(?<!\.)\w+\s*(\,\s*\w+\s*)*\=(.*\,)*[\s|\(]*lambda
```

This implementation has more coverage and is more simple than the previous one.

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -140,7 +140,9 @@ COMPARE_TYPE_REGEX = re.compile(
 )
 KEYWORD_REGEX = re.compile(r'(\s*)\b(?:%s)\b(\s*)' % r'|'.join(KEYWORDS))
 OPERATOR_REGEX = re.compile(r'(?:[^,\s])(\s*)(?:[-+*/|!<=>%&^]+|:=)(\s*)')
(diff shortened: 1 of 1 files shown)


asottile:
I don't think this is solvable with a regex -- it needs to have a parse tree probably -- just a counter example from your proposed regex:

```python
x = f(1, (lambda: None))
```

Sam-Cleveland:
Ok. Working on a solution.

asottile (approved):

asottile:
oops, didn't mean to approve, was testing a user script

PR #1082, opened 2022-07-14
Updates to E203

lucasmaheo:
Tests seem to pass on my machine.

Minimal changes to pass tests to fix behaviour on E203 (documented on Black, many others).

Related to #373.


asottile:
dupe #1047

PR #1086, opened 2022-07-30
allow `is` and `is not` for type comparisons (E721)

### LINKED ISSUE #830: E721: Naive for modules named `types`

### LINKED ISSUE #1084: E721 regression(s)
so #1041 brought up a lot more instances of E721 and I don't feel confident adding this in a release since it's going to cause a bunch of (imo) unnecessary work

for example, here's something which wasn't flagged before but is now:

```python
assert type(x) is not bytes
```

strictly speaking the lint rule probably _should_ match this, but almost every case I can find with `is` and `is not` is being very explicit about the types being checked

I think the most reasonable thing to do here is to only check `==` and `!=` -- those are the ones that are most likely actual mistakes rather than an explicit check.

the other option is revert #1041 -- which might be the quickest thing to do to not block a release -- unclear what the best approach is here

asottile:
see #1084 and #1041 for more details

resolves #1084 
resolves #830

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -129,8 +129,8 @@ COMPARE_SINGLETON_REGEX = re.compile(r'(\bNone|\bFalse|\bTrue)?\s*([=!]=)'
 COMPARE_NEGATIVE_REGEX = re.compile(r'\b(?<!is\s)(not)\s+[^][)(}{ ]+\s+'
                                     r'(in|is)\s')
 COMPARE_TYPE_REGEX = re.compile(
-    r'(?:[=!]=|is(?:\s+not)?)\s+type(?:\s*\(\s*([^)]*[^ )])\s*\))' +
-    r'|\btype(?:\s*\(\s*([^)]*[^ )])\s*\))\s+(?:[=!]=|is(?:\s+not)?)'
+    r'[=!]=\s+type(?:\s*\(\s*([^)]*[^ )])\s*\))'
+    r'|\btype(?:\s*\(\s*([^)]*[^ )])\s*\))\s+[=!]='
 )
 KEYWORD_REGEX = re.compile(r'(\s*)\b(?:%s)\b(\s*)' % r'|'.join(KEYWORDS))
 OPERATOR_REGEX = re.compile(r'(?:[^,\s])(\s*)(?:[-+*/|!<=>%&^]+|:=)(\s*)')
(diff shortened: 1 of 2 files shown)


pycodestyle.py
```diff
@@ -134,8 +134,10 @@
                                      r'\s*(?(1)|(None|False|True))\b')
 COMPARE_NEGATIVE_REGEX = re.compile(r'\b(?<!is\s)(not)\s+[^][)(}{ ]+\s+'
                                     r'(in|is)\s')
-COMPARE_TYPE_REGEX = re.compile(r'(?:[=!]=|is(?:\s+not)?)\s+type(?:s.\w+Type'
```
FichteFoll, line 137: I can't find the rationale for removing `types.*Type` comparisons in the linked issues. As far as I can tell, those should still be included in the checks, shouldn't they?

asottile: I think those got removed because the name `types` isn't necessarily the stdlib `types` module -- not 100% sure though

asottile: the original patch is here: https://github.com/PyCQA/pycodestyle/pull/1041 -- this is just a revert-revert of that with an additional commit on top

FichteFoll: I suppose the precise reasoning is the following, then:

> This check also breaks many ligimitate cases (such as having your own types module with an Enum class for instance, and comparing a value to the enum member).

which I can relate to. You can, of course, also shadow the built-in `type` function, but that's just not something people do in general and having a custom `types` module is much more likely, although also ending your enum in `Type` is most likely less common.

asfaltboy:
@asottile my apologies for breaking the valid use cases in #1041, and thank you for following up with this fix and addressing the nuisance. 🙇

PR #1095, opened 2022-08-14
Remove universal wheel, python 2 is unsupported

gopackgo90:


asottile:
I use this marker as a "is pure python" -- we use python_requires which is the actual metadata that controls this

PR #1105, opened 2022-09-29
make the example ignore a valid prefix

### LINKED ISSUE #1104: Documentation: Error E41?
The example `~/.config/pycodestyle` file shown in the [Configuration section](https://pycodestyle.pycqa.org/en/latest/intro.html#configuration) of the documentation shows `E41` being ignored, but does not list an `E41` error in the [Error codes](https://pycodestyle.pycqa.org/en/latest/intro.html#error-codes):

```
[pycodestyle]
count = False
ignore = E226,E302,E41
max-line-length = 160
statistics = True
```

asottile:
resolves #1104

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -200,7 +200,7 @@ Example::
 
   [pycodestyle]
   count = False
-  ignore = E226,E302,E41
+  ignore = E226,E302,E71
   max-line-length = 160
   statistics = True
 


kjcole:
E71 is not in the list of error codes either....

asottile:
yes it is, there are several E71x codes

kjcole:
In that case, should the documentation explicitly explain that E71 is a wildcard encompassing all? I had no idea from reading that section that E71 is equivalent to E710 through E719 (and presumably, in some far future Python version, E7100 - E7199, and so on). Not at all obvious to me.

asottile:
did you read the help? it's pretty clear to me

Issue #959, opened 2020-08-12
E741: Inconsistent behavior for lambdas


knkski:
It seems like E741 shouldn't apply to variable names inside of comprehensions, due to the limited scope of such variables. If that's the desired behavior anyways, this behavior seems inconsistent to me:

```python
# flagged as lint
foo = [l for l in lines if l]

# not flagged as lint
foo = list(filter(lambda l: l, lines))
```

asottile:
I believe it's just a mistake that this isn't handled for lambdas

the rule is correct for the list comprehension due to the ability to confuse:


```python
foo = [l for l in lines if l]
foo = [1 for l in lines if 1]
```

pktiuk:
I have this issue, but in my case it is clearly inconsistent:

```python
COMP_OPERATIONS = {
    "==": lambda l, r: l == r,   # This is falgged
    ">=": lambda l, r: l >= r,   # This is falgged
    "<=": lambda l, r: l <= r,   # This is falgged
    ">": lambda l, r: l > r,  # This is not flagged
    "<": lambda l, r: l < r,  # This is not flagged
}
```

asottile:
@pktiuk yeah it's not enabled for lambdas at all at the moment -- it's detecting `l >= r` as an assignment (?)

Issue #935, opened 2020-05-15
E741: regression for `re.I` as a default parameter


asottile:
```python
def f(x=re.I):
    pass
```

2.5.0:

```console
$ python -m pycodestyle t.py
$
```

2.6.0:

```console
$ python -m pycodestyle t.py
t.py:1:12: E741 ambiguous variable name 'I'
```

originally reported [here](https://twitter.com/jugmac00/status/1261314387090059266)

regression appears to be introduced by b8db33dfa6cd4029b75efd1e7cab7c1b64d03bc7 via `git bisect` CC @rettenbs

PR #1106, opened 2022-10-07
E741 should work with lambdas

dannysepler:
Closes https://github.com/PyCQA/pycodestyle/issues/959
Closes https://github.com/PyCQA/pycodestyle/issues/935

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -1494,14 +1494,18 @@ def ambiguous_identifier(logical_line, tokens):
     E741: I = 42
 
     Variables can be bound in several other contexts, including class
-    and function definitions, 'global' and 'nonlocal' statements,
-    exception handlers, and 'with' and 'for' statements.
+    and function definitions, lambda functions, 'global' and 'nonlocal'
+    statements, exception handlers, and 'with' and 'for' statements.
     In addition, we have a special handling for function parameters.
 
     Okay: except AttributeError as o:
     Okay: with lock as L:
     Okay: foo(l=12)
+    Okay: foo(l=I)
     Okay: for a in foo(l=12):
+    Okay: lambda arg: arg * l
+    Okay: lambda a=l[I:5]: None
+    Okay: lambda x=a.I: None
     E741: except AttributeError as O:
     E741: with lock as l:
     E741: global I
@@ -1510,17 +1514,23 @@ def ambiguous_identifier(logical_line, tokens):
     E741: def foo(l=12):
     E741: l = foo(l=12)
     E741: for l in range(10):
+    E741: [l for l in lines if l]
+    E741: lambda l: None
+    E741: lambda a=x[1:5], l: None
+    E741: lambda **l:
+    E741: def f(**l):
     E742: class I(object):
     E743: def l(x):
     """
-    is_func_def = False  # Set to true if 'def' is found
+    is_func_def = False  # Set to true if 'def' or 'lambda' is found
     parameter_parentheses_level = 0
     idents_to_avoid = ('l', 'O', 'I')
     prev_type, prev_text, prev_start, prev_end, __ = tokens[0]
-    for token_type, text, start, end, line in tokens[1:]:
+    for index in range(1, len(tokens)):
+        token_type, text, start, end, line = tokens[index]
         ident = pos = None
         # find function definitions
-        if prev_text == 'def':
+        if prev_text in {'def', 'lambda'}:
             is_func_def = True
         # update parameter parentheses level
         if parameter_parentheses_level == 0 and \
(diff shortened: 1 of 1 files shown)


sigmavirus24 (commented):
On my phone and can't test, what happens with `lambda r, l`?

dannysepler:
woops, good catch!

pycodestyle.py
```diff
@@ -1521,6 +1526,10 @@ def ambiguous_identifier(logical_line, tokens):
         # find function definitions
         if prev_text == 'def':
             is_func_def = True
+        if prev_text == 'lambda':
+            is_lambda_arg = True
+        if is_lambda_arg and prev_text == ":":
+            is_lambda_arg = False
```
asottile, line 1532: this _technically_ isn't precise enough due to lambda default arguments:

```pycon
>>> x = [1, 2, 3]
>>> y = lambda a=x[1:]: a
>>> y()
[2, 3]
```

dannysepler: good callout! how bout now?

dannysepler: Also, along these lines there's a false positive in pycodestyle now of:

```python
def a(x=l[1:5]):
    ...
```

i fixed it for lambdas in this diff, but we could make an issue / fix it for function defs after?

pycodestyle.py
```diff
@@ -1545,7 +1554,7 @@ def ambiguous_identifier(logical_line, tokens):
                 ident = text
                 pos = start
         # function parameter definitions
-        if is_func_def:
+        if is_func_def or (is_lambda_def and prev_text in ('lambda', ',')):
```
asottile, line 1557: I think this is still not quite right? `lambda x=a.I: None`

dannysepler: trying to follow -- that example shouldn't trigger an error for `a.I`, right? in the current implementation it wouldn't because the prior token isn't `lambda` or `,`

asottile: oh hmmm, then how about `lambda *i: None` or `lambda **i: None` (I think this already properly handles `lambda *, i: None` and `lambda /, i: None` probably)

pycodestyle.py
```diff
@@ -1509,18 +1512,28 @@ def ambiguous_identifier(logical_line, tokens):
...
     is_func_def = False  # Set to true if 'def' is found
+    is_lambda_def = False
     parameter_parentheses_level = 0
     idents_to_avoid = ('l', 'O', 'I')
     prev_type, prev_text, prev_start, prev_end, __ = tokens[0]
-    for token_type, text, start, end, line in tokens[1:]:
+    for index in range(1, len(tokens)):
+        token_type, text, start, end, line = tokens[index]
+        next_token = tokens[index + 1] if index < len(tokens) - 1 else None
+        next_text = next_token[1] if next_token else None
```
asottile, line 1530: idk if it makes much performance difference here -- but I suspect that these should be inside the `if` below such that they aren't computed unnecessarily -- `pycodestyle`'s checks are fairly "hot" in terms of execution time

pycodestyle.py
dannysepler: https://github.com/PyCQA/pycodestyle/issues/1114

pycodestyle.py
```diff
@@ -1544,11 +1557,17 @@ def ambiguous_identifier(logical_line, tokens):
...
-                ident = text
-                pos = start
+        # function / lambda parameter definitions
+        if is_func_def or is_lambda_def:
+            next_token = tokens[index + 1] if index < len(tokens) - 1 else None
+            next_text = next_token[1] if next_token else None
+            if (
+                next_text and next_text in ':,=)' and
+                prev_text in ('lambda', ',', '*', '**', '(')
+            ):
```
asottile, line 1567: I would write this as:

```python
if index < len(tokens) - 1 and tokens[index + 1][1] in ... and ...:
```

that way you take advantage of short circuiting and don't do unnecessary work

dannysepler: oh yeah, that's simpler too. nice

asottile (approved):
I simplified things a little bit

dannysepler:
Oh nice, thanks so much! 🎉

dhomeier:
2.10.0 now still reports on a construction like `sorted(..., key=lambda l: l.order)`, which 2.9.1 did not.
Is this intentional?

asottile:
yes, see the linked issues as well as the tests in this PR

dhomeier:
I probably misunderstood the comment about `lambda r, l` and why `if l<12:` was `Okay`; but thanks for confirming!

asottile:
the rule is for definitions and declarations, not usages

dhomeier:
You mean in the second case `l` should have never been defined, but once it exists it's ok to test against it?
Seems a weird example then.

asottile:
maybe you're not thinking about this the right way

```python
def f(x):
   l = 1

  if l == 2:
      print(l)
      return l + 5
```

if you only require someone to exclude the lint rule on definitions you'd only have to put `# noqa` on one line, otherwise you'd have to put it on 4 of them -- in many cases you have no control over external libraries so there's no reason to punish `external_library(l=1)` when the other library made a poor choice in naming

asottile:
anyway, this is all off topic so I'm going to stop replying

Issue #1107, opened 2022-10-08
Remove W601, W602, W603, and W604


dannysepler:
These seem to be syntax errors in Python3, so is there much use to lint for them anymore?

asottile:
yeah I think these can be removed since we only target python 3+ now

PR #1111, opened 2022-10-18
Delete W601 -> W604 warnings

dannysepler:
Closes https://github.com/PyCQA/pycodestyle/issues/1107

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -71,11 +71,9 @@ Example usage and output
   optparse.py:69:11: E401 multiple imports on one line
   optparse.py:77:1: E302 expected 2 blank lines, found 1
   optparse.py:88:5: E301 expected 1 blank line, found 0
-  optparse.py:222:34: W602 deprecated form of raising exception
   optparse.py:347:31: E211 whitespace before '('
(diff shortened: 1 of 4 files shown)


docs/intro.rst
```diff
@@ -415,14 +411,6 @@ This is the current list of error and warning codes:
 +------------+----------------------------------------------------------------------+
 | **W6**     | *Deprecation warning*                                                |
 +------------+----------------------------------------------------------------------+
-| W601       | .has_key() is deprecated, use 'in'                                   |
-+------------+----------------------------------------------------------------------+
-| W602       | deprecated form of raising exception                                 |
-+------------+----------------------------------------------------------------------+
-| W603       | '<>' is deprecated, use '!='                                         |
-+------------+----------------------------------------------------------------------+
-| W604       | backticks are deprecated, use 'repr()'                               |
```
FichteFoll, line 424: Perhaps it should still be mentioned in the documentation that these 4 codes have been in use before and should not be used again for future checks?

asottile: I would just remove them as you've done

asottile (approved):

PR #1118, opened 2022-11-01
Fix false positive with E741

dannysepler:
Noted from https://github.com/PyCQA/pycodestyle/issues/959#issuecomment-853816244 -- we needed to limit this check to assignment operators specifically


pycodestyle.py
```diff
@@ -1504,13 +1504,15 @@ def ambiguous_identifier(logical_line, tokens):
...
     Okay: lambda x=a.I: None
+    Okay: if l >= 12:
     E741: except AttributeError as O:
     E741: with lock as l:
     E741: global I
     E741: nonlocal l
     E741: def foo(l):
     E741: def foo(l=12):
     E741: l = foo(l=12)
+    E741: while l := thing:
```
asottile, line 1515: you'll have to put this in the versioned test file rather than the docstring

dannysepler: wow, super speedy. you're right, one sec

asottile: I don't think you need the condition you've added -- `:=` will never appear as a token in 3.7

dannysepler: 100%, i was barking up the wrong tree

testsuite/python38.py
```diff
@@ -53,3 +53,5 @@ def f3(
 #: E221:1:6 E221:1:19
 if (x  := 1) == (y  := 2):
     pass
+#: E741
+while l := thing:
```
asottile, line 57: should probably make this valid syntax as well while you're at it

asottile (approved):

PR #1163, opened 2023-07-15
combine whitespace around operator checks

asottile:
this slightly simplifies some checking


asottile:
I'll be using this to make `'f{x=}'` not broken in 3.12

PR #1174, opened 2023-07-18
port --testsuite to pytest

asottile:


testing/data/W29.py
```diff
@@ -14,8 +14,6 @@ class Foo(object):
 #: W191 W292 noeol
 if False:
 	pass  # indented with tabs
-#: W292:1:36 noeol
-# This line doesn't have a linefeed
```
asottile, line 18: there was a bug here!  in python 3.8 this was reported 3 times.  I didn't bother fixing the bug since it'll be "magically fixed" in python3.9+

I moved the test to its own file

testsuite/test_api.py
```diff
@@ -26,8 +26,8 @@ def setUp(self):
         self._saved_stdout = sys.stdout
         self._saved_stderr = sys.stderr
         self._saved_checks = pycodestyle._checks
-        sys.stdout = PseudoFile()
-        sys.stderr = PseudoFile()
+        sys.stdout = io.StringIO()
+        sys.stderr = io.StringIO()
```
FichteFoll, line 30: FYI, pytest already captures stdout and provides a way to access it: https://docs.pytest.org/en/6.2.x/capture.html#accessing-captured-output-from-a-test-function

asottile: yeah I'm a pytest core dev

asottile: it's a larger refactor and this was already large enough

PR #1179, opened 2023-07-21
fix some wrong renames from testsuite to testing/data vs test

lev-blit:
these changes were initially made in #1175 and I noticed some were made incorrectly (tests instead of testing/data and vice versa)

the second change (with `lost/E11.py`) doesn't really exist in either tests or testing but thought it would be better for it to be tests/lost instead


tests/test_shell.py
```diff
@@ -194,7 +194,7 @@ def test_check_diff(self):
         ))
 
         # no matching file in the diff
-        diff_lines[3] = "+++ b/testing/lost/E11.py"
```
asottile, line 197: this one is correct -- it's meant to be a replacement for `testing/data/E11.py` but "list"

lev-blit: well in [this diff from #1175](https://github.com/PyCQA/pycodestyle/pull/1175/files#diff-e7f1449d934693a05ced6460a039ff825dbc3acb427dce2cda880ca04cc771c4L197-R197) you can see it was `lost` even before your change...

should I still change it to be `list` instead of `lost` -- or did I misunderstand your comment?

asottile: phone typo -- it's supposed to be lost

lev-blit: oh okay. also yes I see what you mean about testing, will remove this part :)

PR #1223, opened 2023-12-28
Adds support for pyproject.toml for python 3.11+ projects

scuml:
PEP 518 and PEP 621 have standardized the use of pyproject.toml for specifying build system requirements and project metadata.  Now that python 3.11 has a toml parsing library, and especially because it's baked right into configparser, it is the right time to give the python community what it has been asking for, and include pyproject as a configuration option.

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -200,7 +200,8 @@ Example::
   max-line-length = 160
   statistics = True
 
-At the project level, a ``setup.cfg`` file or a ``tox.ini`` file is read if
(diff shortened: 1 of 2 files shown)


sigmavirus24:
See #813. We're not adding this support. Maybe I would reconsider if 3.11 were the minimum version we support given it's baked into `configparser` but for now, we can't rely on that and people will be confused why it doesn't work on different versions of Python (or has different behaviour on different versions).

PR #1234, opened 2024-03-08
E502: Fix E502 being disabled after a comment in a logical line

augustelalande:
Since the comment flag was being set and then never disabled in the `explicit_line_join` implementation, any redundant backslash which occured after the first comment in a logical line would go undetected. e.g.
```
y = (
    2 + 2  # \
    + 3  # \
    + 4 \
    + 3
)
```
would not detect the redundant backslash after the 4.


asottile (approved):

augustelalande:
Is anything blocking this?

PR #1237, opened 2024-03-23
E403: add report for module-level dunder names that are misplaced.

### LINKED ISSUE #615: Check if dunder variables are defined *after* imports
[PEP 8](https://www.python.org/dev/peps/pep-0008/#module-level-dunder-names) mandates that dunder variables be specified after docstring and `__future__` imports but before other imports (or other code). In #523, the old check was relaxed not to complain in the case that PEP 8 endorses, but pycodestyle should additionally complain if this is exactly *not* the case.

(via https://github.com/PyCQA/pycodestyle/pull/523)

jayaddison:
According to the [PEP8 Style Guide](https://peps.python.org/pep-0008/#module-level-dunder-names), dunder names at the module level should be placed before any `import` statements, with the exception of any `__future__` module imports.

There should be a single line of spacing before any module-level dunder names, and a single line of spacing after them; they may be grouped.

Related:
 * Resolves #615
 * https://bugs.python.org/issue27187


asottile:
I don't think this should be added.  this is a bad suggestion from pep8

PR #1247, opened 2024-06-15
add new error E204 for whitespace after decorator @

asottile:


asottile:
decided to make this a separate code in case someone wanted to ignore it -- not that this seems controversial at all 🤷

sigmavirus24 (approved):

PR #1278, opened 2025-03-18
add support for PEP 696 generic defaults

### LINKED ISSUE #1277: E251: false positive for PEP 696 TypeVar defaults
https://peps.python.org/pep-0696/

minimal example:

```python
class C[T: (int, str) = int]:
    pass
```

```console
$ ./venv/bin/flake8 t4.py 
t4.py:1:22: E251 unexpected spaces around keyword / parameter equals
t4.py:1:24: E251 unexpected spaces around keyword / parameter equals
```

asottile:
resolves #1277

### DIFF
--- a/tests/test_E901.py
+++ b/tests/test_E901.py
@@ -23,7 +23,7 @@ def lasting(self, duration=300):
 '''
         errors = errors_from_src(src)
         if sys.version_info < (3, 12):  # pragma: <3.12 cover
-            expected = ['E122:4:1', 'E251:5:13', 'E251:5:15']
+            expected = ['E122:4:1']
         else:  # pragma: >=3.12 cover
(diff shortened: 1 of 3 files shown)


tests/test_E901.py
```diff
@@ -23,7 +23,7 @@ def lasting(self, duration=300):
 '''
         errors = errors_from_src(src)
         if sys.version_info < (3, 12):  # pragma: <3.12 cover
-            expected = ['E122:4:1', 'E251:5:13', 'E251:5:15']
+            expected = ['E122:4:1']
```
asottile, line 26: these were due to the E251 check under flowing (which it no longer does now that we track parens in a stack)

sigmavirus24 (approved):

PR #1290, opened 2025-07-12
micro optimisations

abebus:
I [forked Flake8](https://github.com/abebus/flake8) and hardcoded benchmarking utilities for `pycodestyle`. Below are performance comparisons before and after the changes.  

```bash
flake8 src -q -j=1  
```  
*(Executed inside the Flake8 directory.)*  

#### **Before Changes (3 Sample Runs)**  
<details>

```
{'ambiguous_identifier': 6266.212970376301,
 'bare_except': 1362.7702161729383,
 'blank_lines': 2238.630904723779,
 'break_after_binary_operator': 6683.987189751801,
 'break_before_binary_operator': 9769.095276220976,
 'comparison_negative': 2196.917534027222,
 'comparison_to_singleton': 6174.019215372298,
 'comparison_type': 2575.740592473979,
 'compound_statements': 2771.457165732586,
 'continued_indentation': 4376.821457165733,
 'explicit_line_join': 4076.6613290632504,
 'extraneous_whitespace': 2853.963170536429,
 'imports_on_separate_lines': 661.9295436349079,
 'indentation': 1103.8831064851881,
 'maximum_doc_length': 582.9463570856685,
 'missing_whitespace': 6795.916733386709,
 'missing_whitespace_after_keyword': 3579.263410728583,
 'module_imports_on_top_of_file': 1686.7493995196157,
 'python_3000_invalid_escape_sequence': 5908.88710968775,
 'whitespace_around_comma': 1799.519615692554,
 'whitespace_around_keywords': 4659.007205764612,
 'whitespace_around_named_parameter_equals': 3716.1729383506804,
 'whitespace_around_operator': 4288.75100080064,
 'whitespace_before_comment': 2058.0064051240993,
 'whitespace_before_parameters': 2485.6685348278625}
 
 {'ambiguous_identifier': 5786.9895916733385,
 'bare_except': 1216.453162530024,
 'blank_lines': 1999.759807846277,
 'break_after_binary_operator': 6445.15612489992,
 'break_before_binary_operator': 9439.271417133707,
 'comparison_negative': 2072.3779023218576,
 'comparison_to_singleton': 5855.964771817454,
 'comparison_type': 2494.915932746197,
 'compound_statements': 2532.906325060048,
 'continued_indentation': 4167.574059247398,
 'explicit_line_join': 3900.6004803843075,
 'extraneous_whitespace': 2749.7197758206567,
 'imports_on_separate_lines': 627.7822257806245,
 'indentation': 1021.0968775020016,
[...]

PR #1293, opened 2025-07-17
reduce "python_3000_invalid_escape_sequence" execution time by ~60%

abebus:
[Benchmarks](https://github.com/PyCQA/pycodestyle/pull/1290#issue-3225528348)

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -1627,6 +1627,29 @@ def ambiguous_identifier(logical_line, tokens):
         prev_start = start
 
 
+# https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals
+_PYTHON_3000_VALID_ESC = frozenset([
+    '\n',
+    '\\',
+    '\'',
+    '"',
+    'a',
+    'b',
+    'f',
+    'n',
+    'r',
+    't',
+    'v',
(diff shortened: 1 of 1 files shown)


asottile (approved):

pycodestyle.py
```diff
@@ -1637,41 +1660,27 @@ def python_3000_invalid_escape_sequence(logical_line, tokens, noqa):
...
-    ]
-
     prefixes = []
     for token_type, text, start, _, _ in tokens:
-        if token_type in {tokenize.STRING, FSTRING_START, TSTRING_START}:
+        if (
+                token_type == tokenize.STRING or
+                token_type == FSTRING_START or
+                token_type == TSTRING_START
+        ):
```
Zaczero, line 1669: I am watching your stream VOD and I simply wanted to let you know that the reason the previous solution was very slow is that Python LOAD_CONST optimization applies only to trivial Python objects (with guaranteed no side effects). `tokenize.STRING` may have side effects, which prevents this optimization and makes Python construct the set each time you access it in a conditional. My typical workaround for such performance-oriented cases, without polluting the global scope with lots of sets/dicts, is to simply put them in kwargs of the function that run just once. See this bytecode: https://godbolt.org/z/qoPdYvha3

PR #1312, opened 2026-02-17
Fix spurious E231 for nested f-string format specifiers

### LINKED ISSUE #1241: Spurious E231 for nested format substitutions
For
```python
x = 1
digits = 1
print(f"{x:0.{digits:d}f}")  # admittedly the :d is not needed in this minimal repro, but still valid.
```
pycodestyle 2.11.1 spuriously complains about a missing whitespace after the second colon in the f-string:
```
test.py:3:21: E231 missing whitespace after ':'
```

worksbyfriday:
## Summary

Fixes #1241. When a replacement field appears inside an f-string format spec (e.g. `f"{x:0.{digits:d}f}"`), the `:` in the nested `{digits:d}` is a format specifier colon, not a dict colon. Previously pycodestyle reported a false-positive E231 on it.

The fix tracks `FSTRING_MIDDLE` / `TSTRING_MIDDLE` tokens: when `{` is opened immediately after one of these tokens, it is a nested replacement field inside a format spec. A distinct marker `'F'` (instead of `'{'`) is pushed onto the brace stack, and E231 is suppressed for `:` when the top of the stack is `'F'`.

## Test plan

- Added two test cases to `testing/data/python312.py`:
  - `f"{x:0.{digits:d}f}"` — the exact reproducer from #1241
  - `f'{value:{fill}{align}{width}.{precision}f}'` — multiple nested fields in format spec
- Verified dict colons inside f-strings (`f"{ {'a':'b'} }"`) still correctly trigger E231
- All 770 existing tests pass

### DIFF
--- a/pycodestyle.py
+++ b/pycodestyle.py
@@ -874,9 +874,17 @@ def missing_whitespace(logical_line, tokens):
     prev_text = prev_end = None
     operator_types = (tokenize.OP, tokenize.NAME)
     brace_stack = []
+    _after_fstring_middle = False
     for token_type, text, start, end, line in tokens:
         if token_type == tokenize.OP and text in {'[', '(', '{'}:
-            brace_stack.append(text)
+            # After FSTRING_MIDDLE/TSTRING_MIDDLE, `{` opens a nested
+            # replacement field inside a format spec.  Mark it with 'F'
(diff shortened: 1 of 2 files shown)


ekhumoro:
The currently proposed fix fails with a test case like this:

`f'{x:{y:{z:02}}}'`

PR #1313, opened 2026-02-26
fix: replace 5 bare except clauses with except Exception

haosenwang1018:
Replace bare except clauses with except Exception.


asottile:
this seems to be an extremely braindead agentic pr.  please don't waste maintainers time with this trash

